PR #22, opened 2015-11-25
Minor modifications discussed in the last skype call

VChristiaens:
Hola Carlos,
Here is the pull request regarding the last skype call decisions.
Note that I compared the performance of centroid fitting with a 2dmoffat and a 2dgaussian on one of my datasets, and they give consistently a same centroid value within ~0.05px (which I would further assume is the expected accuracy of this kind of method).
I still have some minor recommendations for vip that we could discuss next Friday when I'll be in Liège.

Saludos,
Valentin


carlgogo:
Hola Valentin,

I'll merge the pull_request since it's based on already discussed and **tested** improvements. Then let's discuss your recommendations next week here in Liege.

Saludos,
Carlos

PR #63, opened 2016-12-01
Fix of a bug in the SDI+ADI mode

jmilou:
In SDI+ADI mode, the cube has 4 dimensions and one of the tests to validate the input parameters fails because it assumes that the first dimension of the cube is the temporal range whereas it is the spectral range (see code below). 
I proposed a fix.

Cheers
Julien


carlgogo:
Hi Julien, thank you for the fix. I'll merge it right away!

jmilou:
You’re welcome ! 
I have at last learnt to create a local copy to be able to suggest changes to your code. I’ll do more in the future now !
Thanks for your great tooL !

Julien

carlgogo:
Please, keep them coming!

PR #84, opened 2017-07-21
Changes to routines creating fake companions.

henry-ngo:
1. Keyword "imlib" added to routines "throughput" and "contrast_curve" so that the user can choose the imlib desired for routines such as "inject_fcs_cube" (this routine previously had this keyword as a parameter but there was no way for the routines in "contrcurve.py" to pass this information on.

Reason: Allow for choices other than opencv.

2. In "throughput", I changed the SNR of fake companions from being hard-coded as 10.0 to a user-defined parameter, "fc_snr". I also added this as a keyword/parameter to "contrast_curve" so that a user can pass it through.

Reason: Allow for more flexibility when injecting fake companions.

### DIFF
--- a/vip/phot/contrcurve.py
+++ b/vip/phot/contrcurve.py
@@ -27,15 +27,15 @@ from ..var import frame_center, dist
 
 
 
-def contrast_curve(cube, angle_list, psf_template, fwhm, pxscale, starphot, 
+def contrast_curve(cube, angle_list, psf_template, fwhm, pxscale, starphot,
                    algo, sigma=5, nbranch=1, theta=0, inner_rad=1, wedge=(0,360),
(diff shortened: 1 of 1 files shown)


carlgogo:
@henry-ngo thanks for the PR. I'll merge the changes. Indeed "imlib" was missing in the throughput and CC functions. Regarding the S/N for measuring the throughput, I think it won't hurt to expose that as a keyword, although in principle (and according to my tests) the throughput should be independent of the S/N of the injected companions.

henry-ngo:
Thanks! Dimitri mentioned that he had hacked his version of VIP to always use SNR=25 but I do agree with you that my most recent tests show very little differences between SNR=10,25,99.

PR #85, opened 2017-07-23
Small bug fix in cube_px_resampling()

VChristiaens:
- I corrected the bug that occurs when running vip.preproc.cube_px_resampling(): "frame_px_resampling() got an unexpected keyword argument 'imlib'". This is due to the latest update of frame_px_resampling (keyword 'imlib' was removed); the call to frame_px_resampling in cube_px_resampling was not modified accordingly.

- I added the optional keyword "threshold" in cube_recenter_gauss2d_fit(), this argument is passed to _centroid_2dg_frame(), and hence to vip.var.fit2d_gaussian().

### DIFF
--- a/vip/pca/pca_local.py
+++ b/vip/pca/pca_local.py
@@ -8,7 +8,6 @@ from __future__ import division, print_function
 
 __author__ = 'C. Gomez @ ULg'
 __all__ = ['pca_adi_annular',
-           'pca_adi_annular2',
            'pca_rdi_annular']
 
 import numpy as np
@@ -186,7 +185,6 @@ def pca_rdi_annular(cube, angle_list, cube_ref, radius_int=0, asize=1,
         return cube_out, cube_der, frame 
     else:
         return frame           
-
 
 
 def pca_adi_annular(cube, angle_list, radius_int=0, fwhm=4, asize=3, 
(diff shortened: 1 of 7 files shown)


VChristiaens:
Update:
I did a couple of additional small changes to the following functions:
- vip.phot.contrast_curve(): the definition of the window size for the savitzky-golay filtering of the radial noise profile is inappropriate for large frames (as it is set as 0.1xlargest radial separation). This can lead to negative noise values when approaching the speckle dominated regime (see example curve attached). Now the window is set to the closest odd number to min(int(2*fwhm), largest radial separation-2).
window_for_savitzky_golay_filter_is_too_wide
- vip.preproc.register_and_center_via_speckles(): I changed some hardco
ded values in order for the function to be compatible with NACO data or other pixel scales NIRC2 data.
- vip.var.fit_2dgaussian(): also changed a hardcoded value which could lead to poor results in the case of thresholding before the gaussian fit.

henry-ngo:
Thanks @VChristiaens, I actually just started to encounter this same issue with my data and had to do a hardcode fix on my version of the code. Didn't have time to make a clean change yet, so I'm glad you did this :)

carlgogo:
@VChristiaens thanks for the PR. I'll merge it.

PR #88, opened 2017-08-04
Added option to customize plot titles

wenhaoxuan:
Hi Carlos,
I made tiny additions to add titles to blob detection, contrast curve, snrmap, and pc optimization plots. The reason is that we will be displaying the results of my NIRC2vortex pipeline on a web server.


carlgogo:
Hi @wenhaoxuan, I'll merge this PR. In you need to customize further the plots, please merge all the matplotlib related instructions in one argument/parameter. It could be a dictionary where one could pass the title, size, color and other parameters of the plot.

wenhaoxuan:
Hi @carlgogo, thank you, and will do!

PR #89, opened 2017-08-16
Implemented SSIM metric to calculate distances between frames/cubes

henry-ngo:
The implementation of the SSIM metric appeared to be incomplete (i.e. the code allows the user to select SSIM but it doesn't do anything). I've added the import command to use scikit-image's implementation of the SSIM calculation (it returns the *mean* SSIM between two frames). I've set the default parameters to match the Wang et al. 2004 paper cited in the current description of the SSIM. I also added a clarification in the comments to say that the *mean* SSIM is returned (unlike the other metrics, the SSIM is a matrix, not just a single value).

This calculation is done via scikit-image rather than OpenCV because there's no builtin function in OpenCV to calculate the SSIM. scikit-image is still *very* fast since no rotations are necessary (just a few convolutions). However, if an OpenCV approach is desired in the future, here is how to create an OpenCV version out of OpenCV functions: http://docs.opencv.org/2.4/doc/tutorials/gpu/gpu-basics-similarity/gpu-basics-similarity.html

### DIFF
--- a/vip/stats/distances.py
+++ b/vip/stats/distances.py
@@ -14,27 +14,32 @@ import numpy as np
 import scipy.stats
 from matplotlib import pyplot as plt
 from ..var import get_annulus
+from skimage.measure import compare_ssim as ssim
 
 
-def cube_distance(array, frame, mode='full', dist='sad', inradius=None, 
+def cube_distance(array, frame, mode='full', dist='sad', inradius=None,
                   width=None, plot=True):
     """ Computes the distance (or similarity) between frames in a cube, using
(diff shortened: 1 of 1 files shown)


carlgogo:
@henry-ngo thanks for the fix. For the record, it was implemented using skimage, but their API changed (which broke the call from VIP) and I decided to drop the 'SSIM' option (lack of time to investigate what changed with skimage). It seems that I didn't properly update the docstring and therefor the implementation of the SSIM metric appeared as incomplete.

PR #91, opened 2017-08-29
Check for None using "is" instead of equality and one other small change

henry-ngo:
My machine returns an error when e.g. an array is checked for None type using equality instead of "is". i.e. array==None returns an array of True/False values, which doesn't work in "if" statements (asks for "any" or "all"). It seems like the PEP recommendation for NoneType checking is to use "is" rather than equality. So, I changed the few cases of this. https://www.python.org/dev/peps/pep-0008/#programming-recommendations

Other small change: handle output from the changed fit_2dgaussian in VIP 0.7.5 correctly.


carlgogo:
@henry-ngo thanks for the fixes. Merging.

henry-ngo:
Thanks!

PR #92, opened 2017-08-31
Small fixes

wenhaoxuan:
1) Updated the old pandas sort() function to sort_values() in pca_fullfr.py. This function was changed back in pandas version 0.17.0: http://pandas.pydata.org/pandas-docs/version/0.19.2/generated/pandas.DataFrame.sort_values.html#pandas.DataFrame.sort_values
2) In snr.py added paramater getfig = True in the snrmap function to suppress show() in pp_subplots(). This is to expedite reduction process for the pipeline.
3) Added precaution to avoid the step param in grid() from being zero in pca_fullfr.py
4) The fit_2dgaussian function was changed to output a pandas dataframe, when full_output = True. Must use 1 variable to catch the dataframe, instead of 6 variables.
5) Unindented lines in vip/pca/pca_fullfr, where the opt_pc plot is saved. Those lines were wrongly indented so it only saves for the clever grid, not the manual grid.

### DIFF
--- a/vip/pca/pca_fullfr.py
+++ b/vip/pca/pca_fullfr.py
@@ -838,9 +838,9 @@ def pca_optimize_snr(cube, angle_list, (source_xy), fwhm, cube_ref=None,
             ax2.grid('on', 'major', linestyle='solid', alpha=0.2)
             #plt.savefig('figure.pdf', dpi=300, bbox_inches='tight')
             print()
-            # Optionally, save the contrast curve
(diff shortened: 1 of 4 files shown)


carlgogo:
Hi @wenhaoxuan this PR has conflicts with master. Please check that.

wenhaoxuan:
Hi @carlgogo, sorry about that. It is good now.

PR #105, opened 2017-11-29
Small bug fix: pca_annulus function in utils_pca.py

gruane:
svd_wrapper needs to be imported for the pca_annulus function.
Added the line "from .svd import svd_wrapper" to the top of utils_pca.py.


carlgogo:
Thanks @gruane

PR #106, opened 2017-11-29
Sphere IFS integration - utils_pca

arainot:
Part 2 of SPHERE IFS integration: utils_pca.py update

### DIFF
--- a/vip_hci/negfc/simplex_fmerit.py
+++ b/vip_hci/negfc/simplex_fmerit.py
@@ -12,31 +12,31 @@ from ..var import frame_center
 from ..pca import pca_annulus
 
 
-def chisquare(modelParameters, cube, angs, plsc, psfs_norm, fwhm, annulus_width,  
-              aperture_radius, initialState, ncomp, cube_ref=None, 
+def chisquare(modelParameters, cube, angs, plsc, psfs_norm, fwhm, annulus_width,
+              aperture_radius, initialState, ncomp, cube_ref=None,
               svd_mode='lapack', scaling=None, fmerit='sum', collapse='median'):
     """
     Calculate the reduced chi2:
     \chi^2_r = \frac{1}{N-3}\sum_{j=1}^{N} |I_j|,
-    where N is the number of pixels within a circular aperture centered on the 
+    where N is the number of pixels within a circular aperture centered on the
     first estimate of the planet position, and I_j the j-th pixel intensity.
-    
+
     Parameters
-    ----------    
+    ----------
     modelParameters: tuple
         The model parameters, typically (r, theta, flux).
     cube: numpy.array
         The cube of fits images expressed as a numpy.array.
     angs: numpy.array
-        The parallactic angle fits image expressed as a numpy.array. 
+        The parallactic angle fits image expressed as a numpy.array.
     plsc: float
         The platescale, in arcsec per pixel.
     psfs_norm: numpy.array
-        The scaled psf expressed as a numpy.array.    
+        The scaled psf expressed as a numpy.array.
     fwhm : float
         The FHWM in pixels.
     annulus_width: int, optional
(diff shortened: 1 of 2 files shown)


carlgogo:
Hi @arainot, glad you saw that you didn't need to put back an old version of the ``svd_wrapper`` function. On the other hand, could you clarify where exactly (in which line of which function) we call ``pca_annulus`` with an input 4d cube? My guess is that it comes from the NEGFC.

arainot:
Hi @carlgogo it's actually on line 194 of simplex_fmerit.py in the function: get_values_optimize
I'll send a pull request for that one next.

carlgogo:
@arainot then please include the modified ``simplex_fmerit`` in the PR. Btw, I just spent a few hours fixing some bugs with the previous PR (the injection of companions for ADI+IFS). Please take a look at the changes I had to make. For the future, the idea is that new code must be extensively tested by the submitter/contributor. Take a look here for a quick guide to contributing to open-source projects: http://contribution-guide-org.readthedocs.io/.

arainot:
@carlgogo Sorry I've made the necessary tests on my machine this time. Let me know if there is a problem. I've also uploaded simplex_fmerit.

vip_hci/negfc/simplex_fmerit.py
```diff
@@ -116,60 +116,91 @@ def get_values_optimize(cube, angs, ncomp, annulus_width, aperture_radius,
...
+        msg = 'The annulus and/or the circular aperture used by the NegFC falls '
+        msg += 'outside the FOV. Try increasing the size of your frames or '
+        msg += 'decreasing the annulus or aperture size'
+        if r_guess>centx_fr-halfw or r_guess<=halfw:
+            raise RuntimeError(msg)
+
+        pca_res = np.zeros_like(cube[0])
+
+        for i in range(cube[0].shape):
+            pca_res[i] = pca_annulus(cube[i], angs, ncomp, annulus_width, r_guess, cube_ref,
```
carlgogo: @arainot In this line we are slicing the 4D cube, which means that we're passing a 3D cube to ``pca_annulus``. Isn't it? Then I come back to my previous question, why we need ``pca_annulus`` to handle 4D cubes?

arainot: @carlgogo Yes indeed that's weird, I'll take that off and put it back in a PR if I encounter the need for it later.

carlgogo: @arainot So did you confirm that with the old version of ``pca_annulus`` (which only handles 3D arrays) you could call ``get_values_optimize`` with a 4D cube successfully? If yes, then I don't think I have other comments

arainot: @carlgogo Yes it does work, because we call pca_annulus in 3D from the 4D case. I have just tested if we could do directly pca_annulus in 4D and it does not seem to work. There is an issue with the matrices going from 3D to 4D. Would you like me to remove it or do you want to have a look at it and maybe find a fix? 
I was not able to find the source.

vip_hci/negfc/simplex_fmerit.py
```diff
@@ -46,63 +46,63 @@ def chisquare(modelParameters, cube, angs, plsc, psfs_norm, fwhm, annulus_width,
...
         r, theta, flux = modelParameters
     except TypeError:
         print('paraVector must be a tuple, {} was given'.format(type(modelParameters)))
 
     # Create the cube with the negative fake companion injected
     cube_negfc = cube_inject_companions(cube, psfs_norm, angs, flevel=-flux,
-                                        plsc=plsc, rad_dists=[r], n_branches=1,
-                                        theta=theta, verbose=False)       
-                                      
+                                 plsc=plsc, rad_dists=[r], n_branches=1,
```
carlgogo: Here, spacing!

PR #110, opened 2017-12-13
simplex optimisation in 4D

arainot:
@carlgogo As discussed, here's a clean PR for the simplex optimisation. I found out what was wrong with previous PR: my text editor was getting rid of all the random spaces within the code and GitHub was flagging them as changes.


carlgogo:
@arainot Please upload to Gist your notebook with a simple/clear test of this new functionality

arainot:
@carlgogo Sorry I lost track of where that function is called, if I ever encounter it later I'll do another PR. I'll move on to the contrast curves instead.

PR #139, opened 2018-03-12
`fit_2dgaussian` fixes

r4lv:
`var/fit2d.py:22 fit_2dgaussian()` returns `mean_y, mean_x`.

### DIFF
--- a/vip_hci/preproc/recentering.py
+++ b/vip_hci/preproc/recentering.py
@@ -670,18 +670,21 @@ def cube_recenter_dft_upsampling(array, cy_1, cx_1, negative=False, fwhm=4,
         raise TypeError('Input array is not a cube or 3d array')
 
     n_frames, sizey, sizex = array.shape
-    if not isinstance(subi_size, int):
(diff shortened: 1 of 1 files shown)


vip_hci/preproc/recentering.py
```diff
@@ -670,18 +670,21 @@ def cube_recenter_dft_upsampling(array, cy_1, cx_1, negative=False, fwhm=4,
...
     n_frames, sizey, sizex = array.shape
-    if not isinstance(subi_size, int):
-        raise ValueError('subi_size must be an integer')
-    if sizey % 2 == 0:
-        if subi_size % 2 != 0:
-            subi_size += 1
-            print('subi_size is odd (while frame size is even)')
-            print('Setting subi_size to {} pixels'.format(subi_size))
+    if subi_size is None:
+        pass
```
carlgogo: This is not important as it doesn't change the logic, but I'd prefer to replace this by:
`
if subi_size is not None:
`
1 lines instead of 3 :)

PR #149, opened 2018-03-19
separate optimization and subtraction segments (xloci)

r4lv:


vip_hci/var/shapes.py
```diff
@@ -339,9 +339,9 @@ def get_annulus_segments(array, inner_radius, width, nsegm=8, theta_init=0,
     ----------
     array : array_like
         Input 2d array or image.
-    inner_radius : int
+    inner_radius : scalar
```
carlgogo: I'd suggest to follow the convention in VIP: int or float

vip_hci/var/shapes.py
```diff
@@ -339,9 +339,9 @@ def get_annulus_segments(array, inner_radius, width, nsegm=8, theta_init=0,
...
         The inner radius of the donut region.
-    width : int
+    width : scalar
```
carlgogo: Same here

vip_hci/var/shapes.py
```diff
@@ -372,7 +375,7 @@ def get_annulus_segments(array, inner_radius, width, nsegm=8, theta_init=0,
     rad = np.sqrt((xx - cx) ** 2 + (yy - cy) ** 2)
     phi = np.arctan2(yy - cy, xx - cx)
     phirot = phi % (twopi)
-    outer_radius = inner_radius + width
+    outer_radius = inner_radius + (width*optim_scale_fact)
```
carlgogo: Let's start with this implementation of the optimization segments, it's definitively easier to code this way. We might try later on what I've proposed

vip_hci/pca/pca_local.py
```diff
@@ -44,12 +44,13 @@ def pca_adi_annular(cube, angle_list, radius_int=0, fwhm=4, asize=2,
         central circular area is discarded.
     fwhm : float, optional
         Known size of the FHWM in pixels to be used. Deafult is 4.
-    asize : int, optional
+    asize : scalar, optional
```
carlgogo: Same as before

vip_hci/leastsq/leastsq.py
```diff
@@ -23,9 +23,6 @@
 from ..conf.utils_conf import eval_func_tuple as EFT
 
 
-array = None
```
carlgogo: What's the purpose of this change? The idea of having a global array is that there's no need to pass it in internal function calls. This is more performant when running in multiprocessing mode. Could you please do proper timings and check with both variants: with the global array and passing the big cube every time?

vip_hci/leastsq/leastsq.py
```diff
@@ -54,17 +54,19 @@ def xloci(cube, angle_list, fwhm=4, metric='manhattan', dist_threshold=50,
...
         The number of segments for each annulus. When a single integer is given
-        it is used for all annuli.
+        it is used for all annuli. When set to 'auto', the number of segments is
+        automatically determined for every annulus, based on the annulus width.
     nproc : None or int, optional
         Number of processes for parallel computing. If None the number of
         processes will be set to (cpu_count()/2). By default the algorithm works
         in single-process mode.
-    solver : {'lstsq', 'nnls'}, str optional
+    solver : {'lstsq', 'nnls', 'lsq'}, str optional
```
carlgogo: My short test with the least squares imposing bounds on the variables was not conclusive. As it is right now, lines 230-232, it's equivalent to the non negative least squares. If you'd like to investigate what benefits we could obtain with the scipy.optimize.lsq_linear function, please go ahead! Otherwise I'd prefer to avoid having 'lsq' (scipy.optimize.lsq_linear) here

vip_hci/leastsq/leastsq.py
```diff
@@ -163,15 +162,15 @@ def _leastsq_ann(array, indices, ann, n_annuli, fwhm, angles, delta_rot, metric,
...
     if nproc == 1:
+        res = []
         for j in range(n_segments_ann):
             segm_res = _leastsq_patch(array, j, indices, angles, pa_threshold,
                                       metric, dist_threshold, solver, tol)
             res.append(segm_res)
 
     elif nproc > 1:
-        pool = Pool(processes=int(nproc))
+        pool = Pool(processes=nproc)
```
carlgogo: processes argument of Pool takes an intenger. It's safer to cast nproc as an int here

vip_hci/leastsq/leastsq.py
```diff
@@ -215,7 +214,7 @@ def _leastsq_patch(array, nseg, indices, angles, pa_threshold, metric,
     matrix_res = np.zeros((values.shape[0], yy.shape[0]))
     for i in range(n_frames):
         vector = pn.DataFrame(mat_dists_ann[i])
-        vector.columns = ['i']
+        vector.columns = [i]
```
carlgogo: This is not slicing using the loop variable i, the quotes are needed

r4lv: What exactly is the purpose of setting the label of the column to `'i'`? I thought it is for debugging, as the labels show when displaying the DataFrame in jupyter.

carlgogo: To be honest I don't remember why I named the column, perhaps the line:
vector.columns = ['i']
is not needed in the end

vip_hci/leastsq/leastsq.py
r4lv: As long as `cpu_count()` returns an integer, `nproc` is also an integer because of the floor division `//` in new line 113. I propose to remove the redundant casting.

vip_hci/leastsq/leastsq.py
r4lv: removed in [commit fee553c](https://github.com/vortex-exoplanet/VIP/pull/149/commits/fee553ceded56f1edc9c6d49dd2b5fe8e4e2faff)

vip_hci/var/shapes.py
r4lv: changed in [commit 939fdbf](https://github.com/vortex-exoplanet/VIP/pull/149/commits/939fdbff5a7f7b16815a01ef44d683234b49517a)

vip_hci/leastsq/leastsq.py
r4lv: I did a quick benchmark with `nproc=1`, where the new code was slightly faster than the "global" one (12s vs 12.5s). I'll look into it in more detail.

r4lv: The data cube should not be copied in memory when passed as function argument, as also shown by my benchmark: there is no significant performance difference between the two versions (this time with `nproc=None`):

```
# global:
    raw times: 11.1 8.18 8.53 7.88 7.78 7.91 8.01 7.99 8.11 8.47
    1 loops, best of 10: 7.78 sec per loop

# new:
    raw times: 11.4 7.53 7.6 8.21 8.82 7.97 8.42 7.72 7.77 7.96
    1 loops, best of 10: 7.53 sec per loop
```

As we are already passing many parameters to e.g. `_leastsq_patch`, the gain in readability is minimal. In fact, I find the code even more difficult to read when there are some global variables, as the helper functions are not encapsulated any more.
I propose to revert the change here, and to open a new pull request to remove the `global` variables found in the other submodules too (`frdiff/framediff.py`, `llsg/llsg.py`, `madi/adi_source.py`, `preproc/derotation.py`).

vip_hci/var/shapes.py
r4lv: We could discuss this tomorrow. I'm back in the lab.

vip_hci/leastsq/leastsq.py
r4lv: see [commit 101e481](https://github.com/vortex-exoplanet/VIP/pull/149/commits/101e481c16680312080a0f2679ec544c6f9095f2)

r4lv: benchmark script:

``` python
# timing_xloci.py
# for python 3.6+

import sys, os

global MODULEDIR, TUTDIR
MODULEDIR = "/Users/ralf/.local/share/virtualenvs/pyc-JlLL1mk7/src"
TUTDIR = "/Users/ralf/Documents/KIT/UGA/MA/py/r4lv-vip-tutorial"

def test(module, nproc=1):
    sys.path.insert(0, f"{MODULEDIR}/{module}")

    import vip_hci as vip
    print("loading VIP version {} from <VENV>/{}".format(vip.__version__,
          os.path.relpath(os.path.dirname(vip.__file__), MODULEDIR)))



    cube_orig, angs = vip.fits.open_adicube(f"{TUTDIR}/naco_betapic.fits",
                                            verbose=False)

    cube = vip.preproc.cube_recenter_dft_upsampling(cube_orig, 51, 51,
                                                    fwhm=4.6,
                                                    subi_size=5, negative=True,
                                                    full_output=False,
                                                    debug=False, verbose=False)

    fr_xloci_scl2 = vip.leastsq.xloci(cube, angs, fwhm=4.6, nproc=nproc,
                                      verbose=False)


```

**usage**

First, modify the `MODULEDIR` and `TUTDIR` to point to the corresponding directories on your machine. In the `MODULEDIR`, there are 2 subdirectories which contain the entire VIP code, with corresponding changes to the implementation. The module to test can be selected by passing the folder name to the `test()` function. Run the benchmark with:

``` bash
python -m timeit -s 'import timing_xloci as t' -n1 -r10 -v 't.test("vip-hci", None)'
```

and

``` bash
python -m timeit -s 'import timing_xloci as t' -n1 -r10 -v 't.test("vip-hci-global", None)'
```

**results**
``` text
single processor
    old/global
        ❯ pipenv run python -m timeit -s 'import timing_xloci as t' -n1 -r10 -v 't.test("vip-hci-global", 1)' 
        raw times: 16.3 12.9 13.5 13 13.1 12.7 12.8 12.8 13.4 13.1
        1 loops, best of 10: 12.7 sec per loop
    new
        ❯ pipenv run python -m timeit -s 'import timing_xloci as t' -n1 -r10 -v 't.test("vip-hci", 1)'  
        raw times: 15.2 11.9 14.3 14.8 13.2 12.3 12.6 13.3 13.3 12.4
        1 loops, best of 10: 11.9 sec per loop
multi processor
    old/global
        ❯ pipenv run python -m timeit -s 'import timing_xloci as t' -n1 -r10 -v 't.test("vip-hci-global", None)'
        raw times: 11.1 8.18 8.53 7.88 7.78 7.91 8.01 7.99 8.11 8.47
        1 loops, best of 10: 7.78 sec per loop
    new
        ❯ pipenv run python -m timeit -s 'import timing_xloci as t' -n1 -r10 -v 't.test("vip-hci", None)' 
        raw times: 11.4 7.53 7.6 8.21 8.82 7.97 8.42 7.72 7.77 7.96
        1 loops, best of 10: 7.53 sec per loop
```

carlgogo: Thank you for running the tests. I wasn't very detailed when I mentioned that using the global array the performance was better. Two things need to be tested: memory consumption and  running time. There are a couple of problems with your tests: 
1 . Memory is not tested
2. The naco_betapic.fits file is used, which is a toy sequence. It weights ~2MB whereas a normal cube is an order of magnitude larger. Maybe you start to imagine the problems we encounter when using 8 processes each one copying a 2GB array.

For the tests to be significant you'll need a real cube and to use the memory_profiler package (https://pypi.python.org/pypi/memory_profiler). It'll give you both the memory usage (with the -C option to include children processes) and running time. This might help: https://stackoverflow.com/questions/38358881/how-to-profile-multiple-subprocesses-using-python-multiprocessing-and-memory-pro

vip_hci/leastsq/leastsq.py
carlgogo: I see to options: you either leave it with the integer casting or you use the floor division in each function that uses multiprocessing in VIP.

r4lv: I favour the latter... I will create a new pull request once this one is done, as it is not strictly related to the `xloci`.

vip_hci/leastsq/leastsq.py
r4lv: From how I understand function calls and numpy arrays, the cube should never be copied. I'll look into the memory profiler to verify that, thank you for the links!

carlgogo: And please include the results here for the posterity :)

r4lv: Shouldn't we move the entire discussion elsewhere? Issues, new PR?

carlgogo: If the results show the global array brings no benefit then it will deserve a new PR definitively. For the moment we can just have the new test results here and decide. Even better, it would be great to create a notebook/gist and then link it here (easier to read and visualize).

r4lv: I tested the two variants with an already reduced SPHERE/IRDIS datacube of GJ504 (~410MB, 104×1024×1024) and `memory_profiler` — the results show a memory overhead for the function-argument-variant, which also varies in time while for the global-variant the memory usage stays constant (per annulus). Also, the inner/smaller rings are clearly faster with the global variant!
In the graphic, *top* is the global one and *bottom* where I pass the cube as function argument to the multiprocessing worker. The script was run with `nproc=20` on the `calc` server, with 25 annuli.

benchmark result

vip_hci/leastsq/leastsq.py
r4lv: FYI, in `frdiff/framediff.py:225`, you have written
``` python
df = pn.DataFrame(ind)  # sorting using pandas dataframe
df.columns = ['i', 'j']
df = df.sort_values('i')
```

so the column labels are used for sorting! But in this case, no sorting was done.

PR #155, opened 2018-03-27
ENH: frame_center_radon: do not use multiprocessing with `nproc=1`

r4lv:


r4lv:
I compared the results of the old and the new version using the toy cube from the `tests` folder, they match. I added an `int()` in the last commit, I propose to investigate that choice in an other PR.

[tests_pr155.ipynb.zip](https://github.com/vortex-exoplanet/VIP/files/1856283/tests_pr155.ipynb.zip)

PR #158, opened 2018-03-30
Fixing bug in ``psf_norm``

carlgogo:


r4lv:
This could also be added to `tests/test_center.py:46`, `phot/fakecomp.py:334`, `preproc/recentering.py:540+552`?

PR #177, opened 2018-04-20
Update cube_collapse for 4D data

arainot:
Added ability to collapse 4D data by choosing whether the wavelength or angle gets collapsed. Required for SNR map calculations and photometry.


carlgogo:
Could you explain how do you use this for S/N map calculations? I'm not totally sure, this function needs to handle 4d cubes.

arainot:
After I carefully looked at the full code and some tests, it does not seem this is needed for the SNR map anymore, but I haven't yet been able to determine whether the simplex method or the MCMC would still work without this. I will keep you updated.

PR #181, opened 2018-04-24
fix for pip >= 10

r4lv:
`pip.req` moved to `pip._internal.req` in pip 10.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,10 @@
 import os
 import re
 from setuptools import setup
-from pip.req import parse_requirements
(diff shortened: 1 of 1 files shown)


carlgogo:
Is it ok to assume that people will have pip >= 10 in their machines? I wonder what would be the best way to handle this

r4lv:
As I wrapped both versions in a `try` block, it also works on older versions!

carlgogo:
I missed the try/block. Thanks!

PR #182, opened 2018-04-24
refactor `fits` module

r4lv:


vip_hci/fits/fits.py
```diff
@@ -190,20 +191,18 @@ def write_fits(filename, array, header=None, dtype32=True, verbose=True):
...
-        usually saved in float64 precision.
+    precision : numpy dtype, optional
+        Float precision, by default np.float32 or single precision float.
     verbose : bool, optional
         If True prints message.
 
     """
-    if dtype32:
-        array = array.astype('float32', copy=False)
+    array = array.astype(precision, copy=False)
```
carlgogo: Please also change/fix this in HCIDataset.save

r4lv: thanks, missed that!

vip_hci/fits/fits.py
```diff
@@ -90,7 +91,7 @@ def open_adicube(fitsfilename, verbose=True):
     
     Parameters
     ----------
-    fitsfilename : string
+    fitsfilename : string or Path
         Name of the fits file.
```
carlgogo: I think it's worth explaining here the pathlib.Path option ;)

carlgogo: Hi @r4lv, I think you missed this comment. It's just changing the description of the fitsfilename. Same for open_fits

carlgogo: I mean to extend the description of the argument: Name of the fits file or .... object.

PR #183, opened 2018-04-24
filters / fourier transformation: use `real` instead of `abs`

r4lv:
After checking with David, we think it should be `real`. As the transformations we apply in the Fourier space, e.g. Hanning window filter, shouldn't modify the symmetry of the function, the re-transformed image shouldn't contain any imaginary data — the `abs` breaks the image when it contained negative data, though.


carlgogo:
Good, I agree then!

PR #196, opened 2018-05-11
New option in PCA-ADI+SDI in a single step

VChristiaens:
This commit gives the option of cropping the frames (default is True) or not after rescaling in wavelength during the ADI+SDI in a single step.


carlgogo:
@VChristiaens thanks for the PR. I'll accept it and then put in lowercase the new argument to match the API of VIP.

PR #201, opened 2018-05-14
API: auto-set fwhm on `normalize_psf`

r4lv:


vip_hci/hci_dataset.py
```diff
@@ -901,11 +901,16 @@ def normalize_psf(self, fit_fwhm=True, size=None, threshold=None,
             fwhm = self.fwhm
         else:
             fwhm = 'fit'
-        self.psfn = normalize_psf(self.psf, fwhm, size, threshold, mask_core,
+        self.psfn, flx, meas_fwhm = normalize_psf(self.psf, fwhm, size, threshold, mask_core,
```
carlgogo: I'm a bit puzzled, what is the meas_fwhm variable here? The `normalize_psf` returns two  things when `full_output=True`: the normalized PSF and the flux in a FWHM. We could modify `normalize_psf` to return also the FWHM. Then it could also make sense to set the flux as a new attribute (`star_photometry`)

r4lv: `meas_fwhm` is the measured FWHM. I already modified `normalize_psf` to return that `fwhm` as a third value (see changes to `fakecomp.py`).

r4lv: Would a `HCIDataset.flux` attribute already be used somewhere?

carlgogo: I don't see those changes. Yes, the `star_photometry` is used in the contrast curve calculation and would useful for the ROC module

PR #206, opened 2018-05-18
FEAT: helper functions for parallel computing

r4lv:


r4lv:
python 2 does not support `def fkt(*args, NAMED=ARG)`, I will replace it by `def fkt(*args, **kwargs)`...

PR #209, opened 2018-05-18
store HCIDataset to disk

r4lv:
This PR renames the existing `HCIDataset.save` method to `export`, as it just stores the cube and angles to a FITS file.

A new `save` method was introduced which:
- stores the cube, angles and PSF arrays using numpy's efficient `savez` function
- stores the rest of the object using `pickle`.

``` python
dataset = HCIDataset(cube=...)
dataset.save("data1")
```

Once written to disk the HCIDataset object can be restored using

``` python
dataset = HCIDataset.from_file("data1")
```

### DIFF
--- a/vip_hci/hci_dataset.py
+++ b/vip_hci/hci_dataset.py
@@ -1324,8 +1324,13 @@ class HCIDataset:
         """
         Used by ``pickle``. Filters out the attributes listed in
         ``self.SAVE_WITH_NP``.
+
+        Warning: `copy.copy()` also uses this function, so the ``SAVE_WITH_NP``
+        attributes get missing. Use `self.copy()` instead!
         """
-        state = {k:v for k in self.__dict__ if k not in self.SAVE_WITH_NP}
+        # TODO: use `from six import iteritems; for k,v in iteritems(...)` 
+        state = {k:self.__dict__[k] for k in self.__dict__
+                                                  if k not in self.SAVE_WITH_NP}
         return state
 
     @classmethod
--- a/vip_hci/hci_dataset.py
+++ b/vip_hci/hci_dataset.py
@@ -48,6 +48,12 @@ class HCIFrame(object):
         The FWHM associated with this dataset (instrument dependent). Required
         for several methods (operations on the cube).
     """
+
+
+    SAVE_WITH_NP = ["image"]
+        # TODO: `detection_results` is a tuple of two numpy arrays. Can it also
+        # be saved with numpy?
+
     def __init__(self, image, hdu=0, fwhm=None):
         """ HCIFrame object initialization. """
         if isinstance(image, str):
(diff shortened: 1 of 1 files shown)


r4lv:
...I have no idea on how to add an intelligent test case for that one, do you have an idea?

r4lv:
`pathlib` is not in the standard library of python2. I'll change it.

carlgogo:
You could check out the sklearn project (https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/tests/test_base.py) for ideas on how to test the HCIDataset class.

vip_hci/hci_dataset.py
```diff
@@ -1183,13 +1183,13 @@ def rescale(self, scale, imlib='ndimage', interpolation='bicubic',
         self.cube = cube_px_resampling(self.cube, scale, imlib, interpolation,
                                        verbose)
 
-    def save(self, path, precision=np.float32):
-        """ Writing to FITS file. If self.angles is present, then the angles
-        are appended to the FITS file.
+    def export(self, path, precision=np.float32):
```
carlgogo: Then let's call this method `write_fits` to be consisten with the current function names. Would it be possible to do the same, e.g. renaming `save` to `write_fits` and adding the new `save` method in the HCIFrame class?

r4lv: 👍

vip_hci/hci_dataset.py
```diff
@@ -454,7 +549,7 @@ def stats(self, region='circle', radius=5, xy=None, annulus_inner_radius=0,
         _ = frame_quick_report(self.image, self.fwhm, source_xy, verbose)
 
 
-class HCIDataset:
+class HCIDataset(object):
```
carlgogo: Should we inherit from `object` in all classes for Python 2 compatibility?

vip_hci/hci_dataset.py
```diff
@@ -45,6 +48,12 @@ class HCIFrame:
         The FWHM associated with this dataset (instrument dependent). Required
         for several methods (operations on the cube).
     """
+
+
+    SAVE_WITH_NP = ["image"]
```
carlgogo: Why is this attribute name in capital letters? Could it we include in `__init__` or in the method signature where it's used?

vip_hci/hci_dataset.py
r4lv: Yes. In python2, by default all classes are old-style classes. To make them behave like in python3, we have to explicitly specify classes as new-style by inheriting from `object`. E.g. `super()` is only for new-style classes.

vip_hci/hci_dataset.py
```diff
@@ -1215,3 +1314,149 @@ def subsample(self, window, mode='mean'):
...
+
+        for k in np_stuff_dict:
+            data = np_stuff_dict[k]
+            setattr(obj, k, data)
+
+        return obj
+
+
+
+
```
carlgogo: What is the purpose of all these empty lines?

vip_hci/hci_dataset.py
r4lv: I suggest to leave it as it is for now — I created a new `Saveable` super-class which wraps all that redundant `save` related things, and adds support for keras models etc. We could discuss that on the future PR, when the `Saveable` class is ready?

vip_hci/hci_dataset.py
r4lv: That is a visual hint for myself, that there is a function at the bottom of the file which may be moved to another module 😋

vip_hci/hci_dataset.py
```diff
@@ -1183,13 +1282,13 @@ def rescale(self, scale, imlib='ndimage', interpolation='bicubic',
         self.cube = cube_px_resampling(self.cube, scale, imlib, interpolation,
                                        verbose)
 
-    def save(self, path, precision=np.float32):
-        """ Writing to FITS file. If self.angles is present, then the angles
-        are appended to the FITS file.
+    def write_fits(self, path, precision=np.float32):
```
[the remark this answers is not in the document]
r4lv: 🤔 For the future, we could also think of something like `dataset.cube.write_fits(...)`, `dataset.psf.write_fits(...)` etc. For now, needed the `write_fits` when exchanging the cube with IDL code.

r4lv:
I see only 2 TODOs introduced by this PR:

- using the `six` module for dictionary iteration, which is not necessary but would be slightly more memory efficient on python 2.
- `detection_results` is a tuple of two numpy arrays: This is currently saved with `pickle`, which is not as efficient as numpy's `save` for array data. That will be addressed with the `Saveable` super-class.

r4lv:
They are not `FIXME`s, but rather give hints on how to improve things. If you want to, I could put them into my personal todo list.

r4lv:
I was thinking about rather using the FITS file format to store the entire `HCIDataset`, with the PAs, the PSF, the FWHM etc... that would be much more adapted to the type of data we want to store.

PR #223, opened 2018-06-15
Enhance HCIPostProcAlgo

r4lv:


r4lv:
I also added a HCIPostProcAlgo class for PCA and ANDROMEDA, but I'll wait with committing them to this PR to first hear what you think.

vip_hci/hci_postproc.py
```diff
@@ -16,18 +16,50 @@
...
 from .pca import pca
 import pickle
+import numpy as np
 
 
 class HCIPostProcAlgo(BaseEstimator):
     """ Base HCI post-processing algorithm class.
     """
+    THRESHOLDS_5 = [0.5, 1, 1.5, 2, 2.5, 3, 3.5, 4, 4.5, 5]
+    THRESHOLDS_1 = np.linspace(0.1, 0.99, 10).tolist()
```
carlgogo: We must avoid hard-coding detection thresholds.

vip_hci/hci_postproc.py
```diff
@@ -16,18 +16,50 @@
...
+        else:
+            self.dataset = dataset # needed for snr map generation
+            if verbose:
+                #print("self.dataset overwritten with the one you provided.")
+                # -> debug
+                pass
+
+        return dataset
+
+    def get_probmap(self):
```
carlgogo: The ``get_probmap`` is not a general functionality of all model PSF subtraction algos. Why not just creating a ``get_likelihooh_map`` method for ANDROMEDA's HCIAndromeda class?

vip_hci/hci_postproc.py
```diff
@@ -16,18 +16,50 @@
...
+
+    def _get_dataset(self, dataset=None, verbose=True):
+        if dataset is None:
+            dataset = self.dataset
+            if self.dataset is None:
+                raise ValueError("no dataset specified!")
+        else:
+            self.dataset = dataset # needed for snr map generation
+            if verbose:
+                #print("self.dataset overwritten with the one you provided.")
```
carlgogo: Why is the print commented?

vip_hci/hci_postproc.py
```diff
@@ -16,18 +16,50 @@
...
 
+    def store_args(self, kwargs, *skip):
+        # TODO: this could be integrated with sklearn's BaseEstimator methods
+        for k in kwargs:
+            if k == "self" or k in skip:
+                continue
+            setattr(self, k, kwargs[k])
+
+
+    def _get_dataset(self, dataset=None, verbose=True):
```
carlgogo: Good

vip_hci/hci_postproc.py
r4lv: I am not sure if it should be kept or not. There are two ways of providing a dataset

1. through the constructor, that would be the "normal use"
2. through the `run` method, which is used when the same algo with the same parameters should be run on multiple datasets (e.g. ROC curve generation)

When a user supplies a dataset on *both* places, that is where the `print` would output something. I can think of two scenarios where this could happen:

1. The user first initializes a `HCIPostProcAlgo`, tests it on one single injection (by passing a dataset to the constructor), tunes the parameters, and then decides to build a ROC curve. In that case, the `print` statement is useless, as by design the `EvalRoc` overwrites the original dataset by the one with the injections.
2. The user just wants to run an `HCIPostProcAlgo` on one dataset. By mistake he passes a dataset to the constructor *and* to the `run` method. In that case, it would be useful to warn the user that the dataset passed to the constructor is ignored and overwritten.

vip_hci/hci_postproc.py
r4lv: 🤗

vip_hci/hci_postproc.py
carlgogo: No, these hard-coded thresholds shouldn't be there

vip_hci/hci_postproc.py
carlgogo: ``get_detection_map`` sounds like the right name for this method. This could compute S/N maps for PCA, NMF, LOCI and others and return a likelihood map for Andromeda

vip_hci/hci_postproc.py
carlgogo: Let's leave this question open for the now

vip_hci/hci_postproc.py
r4lv: ✓ removed

vip_hci/hci_postproc.py
r4lv: sounds good!

r4lv: ✓ renamed to `get_detection_map`

PR #224, opened 2018-06-15
Documentation

r4lv:


r4lv:
I just removed the `xloci`-commit, it does not belong here.

vip_hci/var/__init__.py
```diff
@@ -9,5 +9,5 @@
 
 from .filters import *
 from .fit_2d import *
-from .shapes import *
 from .plotting import *
+from .shapes import *
```
carlgogo: what is the purpose of this change in ordering?

r4lv: That is to order the imports alphabetically, like in the other submodules.

PR #227, opened 2018-06-15
save HCIDataset as FITS

r4lv:
Notable changes:

- `HCIDataset.save` updated, so the entire dataset is stored inside a single FITS file (including angles, pixel scale, normalized PSF, ...)
- new `HCIDataset.load` to load back such a dataset
- `open_fits` now accepts a `slice=` for subsetting an array (e.g. SPHERE IRDIS creates a `(2, 2, 64, 64)` PSF for a `(2, 104, 1024, 1024)` cube)

### DIFF
--- a/vip_hci/__init__.py
+++ b/vip_hci/__init__.py
@@ -1,5 +1,7 @@
 from __future__ import (absolute_import)
 
+__version__ = "0.9.6"
+
 from . import andromeda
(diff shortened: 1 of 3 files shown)


r4lv:
We decided to store VIP objects as `npz` instead of `fits`, so this PR is obsolete.

PR #230, opened 2018-06-15
cube_inject_companions: return x and y coordinates

r4lv:
With the new `full_output=True` parameter, `cube_inject_companions` returns the x and y coordinates together with the modified array.

This could be interesting for plotting, or for simplifying the EvalRoc logic.

### DIFF
--- a/vip_hci/hci_dataset.py
+++ b/vip_hci/hci_dataset.py
@@ -738,7 +738,7 @@ class HCIDataset:
 
     def inject_companions(self, flux, rad_dists, n_branches=1, theta=0,
                           imlib='opencv', interpolation='lanczos4',
-                          verbose=True):
+                          full_output=False, verbose=True):
         """ Injection of fake companions in 3d or 4d cubes.
 
         Parameters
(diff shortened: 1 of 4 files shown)


vip_hci/metrics/fakecomp.py
```diff
@@ -184,7 +187,10 @@ def cube_inject_companions(array, psf_template, angle_list, flevel, plsc,
                           '({:.2f} pxs)'.format(posx, posy, rad_arcs,
                                                 rad_dists[i]))
 
-    return array_out
+    if full_output:
+        return array_out, posx, posy
```
carlgogo: This is only returning the x,y position of the last companion injected. The coordinates need to be arranged in a nested tuple perhaps.

r4lv:
In my recent changes, only the positions of the injections in the **first frame and wavelength** are returned, so e.g.

``` python
cube_inject_companions(..., n_branches=2, rad_dists=[1, 2])
```

would return 4 pairs of `(y,x)`, in both the 3D and 4D cases.

Would it be interesting to have the coordinates of the other frames as well?

vip_hci/hci_dataset.py
```diff
@@ -773,6 +773,12 @@ def inject_companions(self, flux, rad_dists, n_branches=1, theta=0,
         verbose : bool, optional
             If True prints out additional information.
 
+        Returns
```
carlgogo: Returning the coordinates should be optional (adding an argument to the method)

carlgogo:
@r4lv what do you mean with "to have the coordinates of the other frames as well"?

r4lv:
If the cube has two frames, and we have 1 injection, the returned value would be `[(y_first_frame, x_first_frame)]`. I was thinking if it would make sense to return something like `[(y_first_frame, x_first_frame), (y_frame_2, x_frame_2)]`?

PR #232, opened 2018-06-16
ANDROMEDA: clean up `shift` submodule

r4lv:


vip_hci/andromeda/shift.py
```diff
@@ -4,7 +4,8 @@
 from __future__ import division, print_function
 
 __author__ = 'Ralf Farkas'
-__all__ = ['calc_psf_shift_subpix']
+__all__ = ['calc_psf_shift_subpix',
+           'subpixel_shift']
```
carlgogo: Why do we need this accesible here?

r4lv: `subpixel_shift` is neither more nor less useful than `calc_psf_shift_subpix`, from the users point of view. But as it is a rather generic function, I would prefer discoverability trough the tab completion and also the documentation.

carlgogo: Then It'd be better to hide these two functions, considering their only use is inside the main ANDROMEDA function.

Issue #226, opened 2018-06-15
Improved verbosity handling


r4lv:
**problem**

For now, multiple output-related *things* exist in VIP:

- `verbose` parameter to functions
- `debug` parameter to functions
- timing information
- progress bars.

**solution**
I propose a new class, `Logger`, which takes care of these. 

Internally, a `Logger` is defined by a *numeric logging level* (roughly following the logging levels of the standard library's `logging` module):

- ERROR = 40
- WARNING = 30
- INFO = 20
- DEBUG = 10

and one or multiple boolean *tags*:

- `progressbar = True`
- `timing = True`.

**example**

The new `Logger` class can be used as follows:

``` python

from xyz import Logger

@Logger.timing()
def andromeda(datacube=None, verbose=False):

	logger = Logger("andromeda", verbose)

	if datacube is None:
		logger.warn("The datacube is not set, that is maybe not what you want!")

	for i in Progressbar(range(1000), verbose=logger):
		logger.debug("loop #{}...".format(i))

	# ...

	logger.info("done.")

```

I can then call `andromeda` like so:

- `andromeda(verbose=True)` / `andromeda(verbose="info")`
	- These set `level=20`, `progressbar=True`, `timing=True`.
	- It will show the timing information, the progressbar, and the output of `info` and `warn` with the string `"andromeda: "` prefixed.
- if I want to disable the timing information, e.g. when I run `andromeda` inside a loop, I can add `-timing` to the `verbose` parameter: `andromeda(verbose="info-timing")` .
	- → `level=20`, `progressbar=True`, `timing=False`
- if I want to disable the output, I use `andromeda(verbose=False)` or `andromeda(verbose='error')`.
	- It sets internally `level=40`, `progressbar=False`, `timing=False`.
- if I want *just*  the progressbar, I can enable it separately using `+progressbar`: `andromeda(verbose='error+progressbar')`
	- → `level=40`, `progressbar=True`, `timing=False`

**notes** 

The `Progressbar` class would check if the `progressbar=True` is set inside the `Logger` object it gets. The `@Logger.timing()` decorator does the same for the `timing` tag.

The `Logger` objects are entirely backwards-compatible to functions which expect a boolean as their `verbose` parameter, as they behave like a boolean when using e.g. `if logger:`. So there is no need to change anything existing.

I have the `Logger` class already coded, except for the `@Logger.timing`. I'll add that shortly, so we can try it out.

What do you think?

PR #236, opened 2018-06-19
Fixing verbosity for ``andromeda``

carlgogo:


r4lv:
Dear Carlos,
could you please check https://github.com/vortex-exoplanet/VIP/issues/226? I would have had an updated version of andromeda with different message handling.

PR #240, opened 2018-06-27
Fake disk injection

jmilou:
First utility function for the fake disk injection...


vip_hci/metrics/fakedisk.py
```diff
@@ -1,25 +1,84 @@
...
+        if np.abs(np.sum(psf)-1) > 1e-4:
+            print('Warning the PSF is not normalized to a total of 1.')
+            print('Normalization was forced.')
+            psf = psf/np.sum(psf)
+        for i in range(nframes):
+            fakedisk_cube[i,:,:] = signal.convolve2d(fakedisk_cube[i,:,:],psf,mode='same')             
+    return fakedisk_cube
 
 
+#import numpy as np
```
carlgogo: The commented lines below seem to be tests, right? They can be removed. Btw, is there a reason why you use pyds9 directly instead of VIP's wrapper?

jmilou:
Hi Carlos, 

Perfect, I will keep going with the rest of the code, thanks for your answer. 

Yes, the commented lines at the end were only for debugging, and I used pyds9 instead of the VIP wrapper because at some point you had removed pyds9 and I got used to using it directly. Now one very useful functionality in the past that made the VIP wrapper superior to pyds9 was the possibility to load cubes of frames. However it is no longer there, so there is not much difference now, from my point of view.

carlgogo:
Great, keep the PRs coming. What do you mean with loading cubes of frames? From a FTIS file stored on disk?

jmilou:
Hi Carlos, 

Here is what I mean by a cube of frames: 
If the variable cube is a 3d cube (for instance a numpy array with cube.shape = (100, 1024, 1024) ), it is very nice to be able to display the cube with

import vip_hci as vip
ds9=vip.fits.ds9()
ds9.display(cube)

This is however no longer possible with the current version of VIP: 
ValueError: too many values to unpack (expected 2)

and we have to load individual frames like that
ds9.display(cube[0,:,:])

Any chance to restore the functionally of loading cubes of frames?

thanks
Julien

r4lv:
I get the same error message on 3D cubes when running `pyds9` directly. It seems to me that this is not related to VIP...

The [development version](https://github.com/ericmandel/pyds9) of pyds9 has support for 3D cubes, so you should install `1.9.dev` through git. The version on pypi is `1.8.1`, which only supports 2D data.

carlgogo:
Hi @jmilou, sending cubes to DS9 (in the same way as it worked in past versions of VIP) works with the development version of pyds9. I mentioned that in the documentation, perhaps without too much detail (https://vip.readthedocs.io/en/latest/#other-dependencies). Solution: to install the development version of pyds9. This should do the trick:

```
pip install --upgrade git+git://github.com/ericmandel/pyds9.git#egg=pyds9
```

PR #242, opened 2018-07-03
FIX: normalize_psf creates wrong shapes

r4lv:
I have a PSF with shape `(39, 39)`, loaded into a `HCIDataset`. Before this patch:

``` pycon
>>> hcidataset.normalize_psf(size=36)
`Force_odd` is True therefore `size` was set to 37
Mean FWHM: 4.801
Flux in 1xFWHM aperture: 1.307
Normalized PSF array shape: (37, 37)
The attribute `psfn` contains the normalized PSF
`fwhm` attribute set to 4.801

>>> hcidataset.normalize_psf(size=36, force_odd=False)
Mean FWHM: 4.801
Subframe size is even (while frame size is odd)
Setting subframe size to 37 pixels
Subframe size is even (while frame size is odd)
Setting subframe size to 37 pixels
Flux in 1xFWHM aperture: 1.307
Normalized PSF array shape: (37, 37)
The attribute `psfn` contains the normalized PSF
`fwhm` attribute set to 4.801
```

While `force_odd=False` tells `normalize_psf` to skip its checks, it is then erroneously passed as `force=force_odd` to the internal call to `frame_crop`.

After the patch, it works as expected:

``` pycon
>>> hcidataset.normalize_psf(size=36, force_odd=False)
Mean FWHM: 4.801
Flux in 1xFWHM aperture: 1.301
Normalized PSF array shape: (36, 36)
The attribute `psfn` contains the normalized PSF
`fwhm` attribute set to 4.801
```

### DIFF
--- a/tests/test_fakecomp.py
+++ b/tests/test_fakecomp.py
@@ -1,0 +1,43 @@
+"""
+
+Tests for `metrics/fakecomp.py`.
+
+"""
+
+from __future__ import division, print_function
+
+import numpy as np
+import pytest
+
+import vip_hci as vip
+
+allclose = np.testing.assert_allclose
+array = np.array
+
+
+@pytest.mark.filterwarnings("ignore:invalid value encountered in true_divide")
+def test_normalize_psf_shapes():
+    """
+    Test if normalize_psf produces the expected shapes.
+    """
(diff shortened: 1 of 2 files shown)


carlgogo:
Thanks @r4lv. Merging this PR now.

PR #247, opened 2018-07-16
ENH: update ANDROMEDA

r4lv:


carlgogo:
CI is broken, please check that out @r4lv

PR #251, opened 2018-07-16
FIX CI: add `plot=False` to tests

r4lv:
Travis CI complains about `no display name and no $DISPLAY environment variable`, as tkinter is used as default backend for matplotlib.

Two possibilities to resolve this issue:

1. force `agg` as a backend (see https://github.com/vortex-exoplanet/VIP/commit/b6758ef45d80c9fa30d2ab4d8c28bc2040b9fcbd)
2. never call `plt.figure`

This PR sets `plot=False` for some functions in `test_center.py`, so no figures are created.

### DIFF
--- a/tests/test_center.py
+++ b/tests/test_center.py
@@ -193,7 +193,7 @@ def test_2d(debug=False):
                                         stddev=1)
 
         method_args = dict(fwhm=1, subi_size=5, model=model, verbose=False,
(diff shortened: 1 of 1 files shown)


r4lv:
This should fix the CI errors in https://github.com/vortex-exoplanet/VIP/pull/247.

PR #256, opened 2018-07-18
Minor change in vip.metrics.fakedisk.create_fakedisk_cube

jmilou:
I added more flexibility in the function create_fakedisk_cube by using the argument psf to specify a FWHM for a quick gaussian convolution. Previous to this change, the argument needed to be the numpy.ndarray to be used by the convolution


carlgogo:
@jmilou you might need to pull and merge/rebase to solve the conflicts.

PR #265, opened 2018-07-31
FIX: replace unicode quotation marks with apostrophes

r4lv:
(only) the file `rescaling.py` contained

``` text
‘	U+2018    	LEFT SINGLE QUOTATION MARK / General Punctuation
’	U+2019    	RIGHT SINGLE QUOTATION MARK / General Punctuation
```

which is unsupported on python2. I replaced those quotation marks with 
``` text
'	U+0027    	APOSTROPHE / Basic Latin
```

to be consistent with the other files.

#### alternative solution

An alternative would be to add `# coding: utf-8` to the top of every file, to allow unicode characters in docstrings/comments etc, but I see no big gain in using e.g. typographic quotes in source code.


r4lv:
The following command finds python files with non-ascii characters in the VIP directory:

``` bash
find . -type f -name "*.py" -print0 | xargs -0 ggrep --color='auto' -P -n "[\x80-\xFF]"
```

(for mac OS with `brew install coreutils`)

PR #272, opened 2018-08-03
split requirements into 2 files

r4lv:


carlgogo:
@r4lv Please check the increased runtime

r4lv:
The runtime seems to have increased because the `requirements` were changed. On the first build with new requirements, `astropy` is built from source, and a wheel is created. On the subsequent builds, that wheel is used, and the build should be faster.

PR #289, opened 2018-08-31
Include the IFS pixel scale

arainot:
Added the choice between IRDIS and IFS pixel scale.


carlgogo:
Thanks, looks good!

arainot:
Whoops sorry, pressed the wrong button haha

PR #293, opened 2018-09-04
Injection of extended features

VChristiaens:
Added a function to inject fake companions along a trace (e.g. a spiral arm) in a raw cube.


carlgogo:
Hi @VChristiaens, thanks for the PR, could you include a short routine script showing how this is used? It could be useful for creating an automated test as well. Thanks.

carlgogo:
Thanks!

PR #281, opened 2018-08-09
[wip] TST: tests for var/shapes.py

r4lv:
> ### **work in progress**, do not merge

Notable changes:

- tests for functions in `var/shapes.py`
- remove `get_annulus_cube`, as it is never used. One should use `get_annulus_segments`.
- remove `get_annulus` in favour of `get_annulus_segments`.
- fix https://github.com/vortex-exoplanet/VIP/issues/279

PR #294, opened 2018-09-05
TST: tests for var/shapes.py (2)

r4lv:
- Refactoring of `shapes.py`, so the functions have a common API
- Tests.

### DIFF
--- a/tests/test_shapes.py
+++ b/tests/test_shapes.py
@@ -170,7 +170,12 @@ def test_get_annulus_segments():
     assert repr(res) == repr(truth)
     # TODO: cannot compare using `allclose`, as elements have variable length!
 
-    # masked arr
+    # tuple as input:
+
+    res = vip.var.get_annulus_segments((6, 6), 2, 3, nsegm=3)
+    assert repr(res) == repr(truth)
+
+    # masked arr:
 
     res = vip.var.get_annulus_segments(arr, 2, 3, mode="mask")[0]
     truth = np.array([[0., 0., 0., 1., 1., 1., 1., 0., 0., 0.],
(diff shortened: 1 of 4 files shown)


r4lv:
There are a few changes to the signature of some functions. I updated the other functions inside VIP as well, but please check if you use them somewhere!

Changed signatures:

- `get_ellipse`
- `get_annulus_segments`
- `get_ell_annulus`
- `get_circle`

Please check the commits marked with `API:` for details.

r4lv:
Ready to be merged 😉

carlgogo:
You've changed the signature of ``get_annulus_segments``? That function is used very often in VIP. Could you elaborate on the changes?

r4lv:
In this PR I only changed the *return type* of `get_annulus_segments`. See the commit message:

> get_annulus_segments with mode="val" returned a ndarray with `nsegm` entries. It now returns a list instead, to match the other two modes.

The *signature* of `get_annulus_segments` was changed in https://github.com/vortex-exoplanet/VIP/pull/281, which was the "first part" of my work on `shapes.py`, which you already merged. There, I replaced the `output_values` parameter with `mode`. ([corresponding lines in the other PR](https://github.com/vortex-exoplanet/VIP/pull/281/files#diff-ea6654e8fbb106ef1d599e124bd134f5L389))

carlgogo:
It's ok as long as the change doesn't affect the logic or break other functions in VIP. That seems to be the case. 

Regarding the new helping function ``_image_or_shape``, it should better be moved to ``conf.conf_utils`` which contains ``check_array``.

r4lv:
I verified the other functions to make sure nothing breaks. The functions which call `get_annulus_segments` with `mode="value"` always work with one single annulus only and call `get_annulus_segments(...)[0]`, so nothing should change there.

PR #307, opened 2018-09-19
FIX: EvalRoc still used old signature of compute_tpr_fps

r4lv:


vip_hci/metrics/roc.py
```diff
@@ -194,11 +194,14 @@ def compute_tpr_fps(self, npix=1, min_distance=1):
...
 
             for m in self.methods:
-                res = compute_binary_map(m.probmaps[i], m.thresholds, x, y,
-                                         npix=npix, min_distance=min_distance)
-                m.detections.append(res[0])
-                m.fps.append(res[1])
-                m.bmaps.append(res[2])
+                dets, fps, bmaps = compute_binary_map(
+                    m.probmaps[i], m.thresholds, fwhm=self.dataset.fwhm,
+                    injections=(y, x), **kwargs
```
carlgogo: All the functions in VIP that are exposed to the user, use the convention (X,Y). I've made this modification to compute_binary_map in a previous PR.

carlgogo:
Travis is failing (for py3.6, py3.7) since my last PR. Not sure what happened, the error is not very informative. Please investigate.

vip_hci/metrics/roc.py
r4lv: 👍

r4lv:
I'll check the CI.

r4lv:
Yesterday (!) the new matplotlib 3.0 was released. It removes a function (`matplotlib.pyplot.hold`) which is used in the `negfc` module. Also, matplotlib 3.0 is python 3 only. Until now, we used 2.2.2.

From the [matplotlib homepage](https://matplotlib.org):

> Matplotlib 3.0 Python 3 only.
> 
> For Python 2 support, Matplotlib 2.2.x will be continued as a LTS release and updated with bugfixes until January 1, 2020.

I think we should fix the matplotlib version in the setup.py to 2.x. We could then think of replacing the concerned functions with ones that are supported in both matplotlib 2 and 3, to remove this restriction in a second step.

carlgogo:
Sounds good to me

PR #313, opened 2018-09-19
REL/DOC: update Sphinx to latest 1.8.0

r4lv:
The optional dependency `pyds9` prevent us from using a recent version of sphinx. Because building the documentation for VIP is important, I made the decision in favour of VIP. This PR

- makes use of the latest Sphinx 1.8.0
- pins Sphinx to exactly 1.8.0, so no matter which machine builds the documentation (various developers, readthedocs, ...), it always results in the same output.

and thus 

- makes the installation of pyds9 even more complicated.

This means now:

``` bash
# downgrade:
pip install sphinx==1.5.6
# install pyds9:
pip install git+https://github.com/ericmandel/pyds9.git#egg=pyds9
# update sphinx again:
pip install -U sphinx
```

This *only* concerns developers, as only developers have installed the `dev` release of VIP. Users do not have Sphinx installed (as it is not listed in VIP's `requirements.txt`), so they can upgrade VIP as they wish and never see the effect of this PR.

I also updated the documentation accordingly.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -144,6 +144,11 @@ users who want to experiment with the code.
 
 Other dependencies
 ^^^^^^^^^^^^^^^^^^
+
+
+OpenCV
+~~~~~~
+
 ``OpenCV`` (Open source Computer Vision) provides fast C++ image processing
 operations and is used by ``VIP`` for basic image transformations. If you don't
 have/want the ``OpenCV`` python bindings (``OpenCV`` is optional since ``VIP``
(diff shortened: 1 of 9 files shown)


r4lv:
@carlgogo, I think you would need to change the configuration in readthedocs, too. In my "fork documentation" I had `requirements.txt` listed under « Admin → Advanced Settings → Requirements File », I changed it to `requirements-dev.txt`, so that it installs/uses the correct Sphinx 1.8.0. The docs for this PR are [here](https://vip-test-.readthedocs.io/en/patch-sphinx-un156/).

The current "official" VIP documentation on RTD is built with 1.7.9 (see bottom of the documentation page). Let's keep an eye on that.

PR #315, opened 2018-09-21
DEV: linter configurations for PEP8

r4lv:
This PR adds the `flake8` style checker to VIP, to help developers keep a consistent code style.

[PEP 8](https://www.python.org/dev/peps/pep-0008) is the style guide for python, but it is no set of rules: it leaves some freedom to the developer. To maintain a consistent style in VIP, it is however necessary to make precise decisions on how to interpret PEP 8. This is achieved by defining a set of *rules* for flake8. I selected the codes so that they match the existing coding style of VIP (as good as possible).

## 1. flake8 plugins

I also added the [flake8-docstrings](https://gitlab.com/pycqa/flake8-docstrings) and [flake8-bandit](https://pypi.org/project/flake8-bandit/) packages to the `requirements-dev.txt`. They add more error codes/checks to flake8:

- `flake8-docstrings` verifies for correct formatting of docstrings according to PEP 247.
- `flake8-bandit` checks the security of the code (execution of untrusted input, etc). Not directly related to VIP, but a good choice in general, I think.

## 2. usage

To make use of the rules defined in this PR, flake8 needs to be configured with the Editor / IDE.

### 2.1 Sublime Text

Through [Package Control](https://packagecontrol.io), install [SublimeLinter](https://github.com/SublimeLinter/SublimeLinter) and [SublimeLinter-flake8](https://github.com/SublimeLinter/SublimeLinter-flake8). The path to the `flake8` executable needs to be set in the `Preferences → Package Settings → SublimeLinter → Settings`. A minimal configuration file would look like this:

``` json
{
	"debug": false,
	"lint_mode": "load_save",
	"linters":
	{
		"flake8":
		{
			"args": [],
			"disable": false,
			"executable": "/usr/local/miniconda3/envs/ma-36/bin/flake8"
		}
	},
}
```

SublimeLinter-flake8 automatically ready the `.flake8` configuration file from the *root* of the project.

### 2.2 Visual Studio Code

Refer to the [documentation](https://code.visualstudio.com/docs/python/linting) on how to use flake8 with VSCode.

TLDR; [python extension](https://marketplace.visualstudio.com/items?itemName=ms-python.python), [set python interpreter](https://code.visualstudio.com/docs/python/python-tutorial#_select-a-python-interpreter), `⇧⌘P → Python: select linter → flake8`

The manual way would be to use the project specific settings file (`.vscode/settings.json`). It should look like this:

``` json
{
    "python.linting.pylintEnabled": false,
    "python.linting.flake8Enabled": true,
    "python.linting.enabled": true,
    "python.pythonPath": "/usr/local/miniconda3/envs/ma-36/bin/python"
}
```

### 2.3 PyCharm

PyCharm has no direct support for flake8. But it has pycodestyle built in, which is the style checker flake8 uses too.

You have to set the ignore codes yourself in `Settings → Editor → Inspections → Python → PEP 8 coding style violation`:
- remove the `W29` code (ignore trailing whitespace), but *keep* the `E501` code (line too long, PyCharm does that itself by another inspection)
- add the codes from the `.flake8` configuration file

The maximum line length should be set to `80` in `Settings → Editor → Code Style → Python → Wrapping and Braces → Hard wrap at`.

`pycodestyle` does not check for errors in the docstrings, like `flake8` does (using the `flake8-docstrings` plugin)


r4lv:
I think we could also add a section `contributing` or `developers` to the VIP documentation, where we could provide a starting point for people who want to help VIP development.

PR #323, opened 2018-09-24
MAINT/FEAT: slim down EvalRoc

r4lv:
This PR factors out some functionalities from the EvalRoc class, making them more accessible.

## 1. HCIDataset and injections

I think the HCIDataset itself could be a good location to keep track of all the injection positions.

Changes:

- `HCIDataset.inject_companions()` creates a `.injections_yx` attribute with all the injections (useful later on for ROC curves)
- `HCIDataset.copy()` which creates a full separate copy of the dataset

 How it could be used in the future:

``` python
# original dataset object:
ds = HCIDataset(cube=...)

# create a new HCIDataset with injections:
dsi = ds.copy()
dsi.inject_companions(flux=1000, rad_dists=[10, 20])
dsi.inject_companions(flux=500, rad_dists=[10, 20], theta=90)  # more injections

# process data cube:
algo = HCIMedianSub(...)
algo.run(dsi)

# make ROC curve out of it:
r = Roc()
r.add(algo.detection_map, dsi.injections_yx)
r.plot()
```

### DIFF
--- a/vip_hci/hci_dataset.py
+++ b/vip_hci/hci_dataset.py
@@ -856,14 +856,12 @@ class HCIDataset(Saveable):
 
         return fake_datasets
 
-
-
-
     def copy(self, deep=True):
         """
         Create an in-memory copy of this HCIDataset.
 
-        This is especially useful when creating a 
+        This is especially useful for keeping a backup copy of the original
+        dataset before modifying if (e.g. with injections).
 
         Parameters
         ----------
--- a/vip_hci/hci_dataset.py
+++ b/vip_hci/hci_dataset.py
@@ -777,8 +777,8 @@ class HCIDataset(Saveable):
         Returns
         -------
         yx : list of tuple(y,x)
-            [full_output] Pixel coordinates of the injections in the first frame
-            (and first wavelength for 4D cubes). These are only the new
+            [full_output=True] Pixel coordinates of the injections in the first
+            frame (and first wavelength for 4D cubes). These are only the new
             injections - all injections (from multiple calls to this function)
             are stored in ``self.injections_yx``.
 
--- a/vip_hci/hci_dataset.py
+++ b/vip_hci/hci_dataset.py
@@ -475,7 +475,7 @@ class HCIDataset(Saveable):
     """
 
     _saved_attributes = ["cube", "psf", "psfn", "angles", "fwhm", "wavelengths",
-                         "px_scale"]
+                         "px_scale", "cuberef"]
 
     def __init__(self, cube, hdu=0, angles=None, wavelengths=None, fwhm=None,
                  px_scale=None, psf=None, psfn=None, cuberef=None):
--- a/vip_hci/hci_dataset.py
+++ b/vip_hci/hci_dataset.py
@@ -480,7 +480,7 @@ class HCIDataset(Saveable):
     """
 
     _saved_attributes = ["cube", "psf", "psfn", "angles", "fwhm", "wavelengths",
-                         "px_scale", "cuberef"]
+                         "px_scale", "cuberef", "injections_yx"]
 
     def __init__(self, cube, hdu=0, angles=None, wavelengths=None, fwhm=None,
                  px_scale=None, psf=None, psfn=None, cuberef=None):
@@ -584,7 +584,7 @@ class HCIDataset(Saveable):
         if self.px_scale is not None:
             print('Pixel/plate scale: {}'.format(self.px_scale))
 
-        self.injections_yx = []
+        self.injections_yx = None
 
     def collapse(self, mode='median', n=50):
         """ Collapsing the sequence into a 2d array.
@@ -802,6 +802,9 @@ class HCIDataset(Saveable):
             rad_dists, n_branches, theta, imlib, interpolation,
             full_output=True, verbose=verbose
         )
+
+        if self.injections_yx is None:
+            self.injections_yx = []
 
         self.injections_yx += yx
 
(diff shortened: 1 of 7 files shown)


r4lv:
## 2. Creating copies of a dataset with random injections

The `EvalRoc.inject_and_postprocess()` was split out into a `cube_copies_with_injections` function, which
- creates copies of a given dataset
- injects different fake companions
- returns a list of the new "fake" data sets.

I also added the same functionality to the OO API, as `HCIDataset.copies_with_injections`, which complements the `.inject_companions` method, but returns a list of new `HCIDataset` objects instead.

vip_hci/hci_dataset.py
```diff
@@ -772,12 +775,13 @@ def inject_companions(self, flux, rad_dists, n_branches=1, theta=0,
         Returns
         -------
         yx : list of tuple(y,x)
-            [if full_output] Pixel coordinates of the injections in the first
-            frame (and first wavelength for 4D cubes).
+            [full_output] Pixel coordinates of the injections in the first frame
```
carlgogo: We haven't discussed thoroughly how to use the squared brackets in the docstrings. It's more informative to put [full_output=True] instead of just [full_output]

vip_hci/hci_dataset.py
```diff
@@ -789,16 +793,101 @@ def inject_companions(self, flux, rad_dists, n_branches=1, theta=0,
...
-                                               full_output=True,
-                                               verbose=verbose)
+        self.cube, yx = cube_inject_companions(
+            self.cube, self.psfn, self.angles, flux, self.px_scale,
+            rad_dists, n_branches, theta, imlib, interpolation,
+            full_output=True, verbose=verbose
+        )
+
+        if not hasattr(self, "injections_yx"):
+            self.injections_yx = []
```
carlgogo: Let's try to avoid creating attributes outside __init__. It's more clear when we have all the attributes in __init__, even if they are initialized with None or something else

vip_hci/hci_dataset.py
```diff
@@ -789,16 +793,101 @@ def inject_companions(self, flux, rad_dists, n_branches=1, theta=0,
...
+                    uses np.random.normal
+
+        Returns
+        -------
+        fake_datasets : list of HCIDataset
+            List containing ``n_copies`` fake datasets.
+
+        """
+
+        fake_datasets = []
```
carlgogo: I'm in favor of this new method. The only problem I see is the memory management. Let's say we have a 10 Gb datasets, and we request 100 copies... I'd suggest to check the available memory as it's done in vip_hci.pca.pca (full-frame PCA)

vip_hci/hci_dataset.py
```diff
@@ -789,16 +793,101 @@ def inject_companions(self, flux, rad_dists, n_branches=1, theta=0,
...
+        """
+        Create an in-memory copy of this HCIDataset.
+
+        This is especially useful when creating a 
+
+        Parameters
+        ----------
+        deep : bool, optional
+            By default, a deep copy is created. That means every (sub)attribute
+            is copied in memory. While this requires more memory, one can safely
```
carlgogo: Again, we should think about memory management

vip_hci/hci_dataset.py
```diff
@@ -789,16 +793,101 @@ def inject_companions(self, flux, rad_dists, n_branches=1, theta=0,
...
+        return fake_datasets
+
+
+
+
+    def copy(self, deep=True):
+        """
+        Create an in-memory copy of this HCIDataset.
+
+        This is especially useful when creating a
```
carlgogo: The sentence is incomplete

vip_hci/hci_dataset.py
r4lv: 👍

vip_hci/hci_dataset.py
r4lv: 👍

vip_hci/hci_dataset.py
r4lv: 👍

vip_hci/hci_dataset.py
r4lv: I just added the memory checking using `conf.mem.check_enough_memory`. However, maybe we could think of this from another perspective: instead of storing all the cubes/datasets in a big list, we could also separate the problem in two steps:

1. Create the information required for the injection (position, flux, ...) and store it in a variable. That could also be stored to disk, to reproduce the injections later.
2. Based on that information, use a generator/iterator, so that only the cube, which is currently processed, is stored in memory.

in code:

``` python
ds = HCIDataset(...)
inj_info = make_injections(angles, n_copies, dist_flux=...)
for dataset in ds.generate_copies_with_injections(inj_info):
    # just this current dataset ist stored in memory, and
    # deleted on the next iteration element
```

I'll think about a simple way to do so.

vip_hci/hci_dataset.py
r4lv: 👍

r4lv:
I just created a copy of the concerned functions, which now use a generator. The cubes/datasets are created "on the fly". The python garbage collector *should* remove the object when the `for` loop reaches the next item. I think that is much cleaner than the memory testing, as we rely on the python interpreter, instead of trying to manage memory ourselves, which is quite complicated.

I kept the "old" version, too, so we can decide which one to remove.

(Note that `del` does, generally, *not* free up memory.)

r4lv:
This PR should be ready. Sorry for the messy *merge commit*... the only change since our last discussion is the removal of the "copied cubes as list" version, so we only keep the generator one.

PR #327, opened 2018-09-26
MAINT: improve ANDROMEDA logging

r4lv:
This PR simplifies the code responsible of logging, using a new `printer()` factory function. It creates a function which can be used instead of `if verbose: print("...".format())`, like so:

``` python
def my_function(param=1, verbose=True):
    info = printer(verbose)

    # only prints when verbose is True:
    info("the user param is {}", param)
```

a more sophisticated example:

``` python
def my_function_2(loops=3, verbose=True, debug=False):
    info = printer(verbose)
    log = printer(debug, prefix="[debug] ")

    info("working with {} loops", loops)
   
    for i in range(loops):
        log("loop {} of {}...", i+1, loops)
```

``` pycon
>>> my_function_2()
working with 3 loops
>>> my_function_2(verbose=True)
working with 3 loops
[debug] loop 1 of 3...
[debug] loop 2 of 3...
[debug] loop 3 of 3...
```

For the other smaller changes please refer to the commit messages below.

### DIFF
--- a/vip_hci/andromeda/andromeda.py
+++ b/vip_hci/andromeda/andromeda.py
@@ -290,8 +290,10 @@ def andromeda(cube, oversampling_fact, angles, psf, filtering_fraction=.25,
 
     if np.asarray(tnd).ndim == 0:  # int or float
         log("Throughput map: Homogeneous transmission: {}%", tnd * 100)
-    else:  # TODO: test if really 2d map?
+    elif np.asarray(tnd).ndim == 2:
         log("Throughput map: Inhomogeneous 2D throughput map given.")
+    else:
+        raise ValueError("Throughput `tnd` should be an float or a 2d map")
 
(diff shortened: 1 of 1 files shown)


r4lv:
I did some checking on how other projects handle output to the user. It seems to me that there is not one single way to do it:

- matplotlib uses python's build-in `logging` module:
	``` python
	import logging
	_log = logging.getLogger(__name__)
	_log.info(...)
	_log.debug(...)
	```
- keras uses plain `if verbose: print(...)`
- tensorflow has it's own logging functions:
	``` python
	from tensorflow.python.platform import tf_logging as logging
	if verbose:
		logging.info(...)
	```
- numpy does not log...

Thus I see no problem in using a custom-tailored solution for VIP.



I updated the code a bit to make it clearer to read.

PR #331, opened 2018-10-15
FIX: wrong indices in cube_drop_frames

### LINKED ISSUE #330: preproc.cube_drop_frames uses unintuitive indices
### current behaviour

`cube_drop_frames` takes two parameters, `n` and `m`:

https://github.com/vortex-exoplanet/VIP/blob/edae4327b9dc3e118133cefb1f8d328c22f27786/vip_hci/preproc/cosmetics.py#L136-L139

and then slices the cube using

https://github.com/vortex-exoplanet/VIP/blob/edae4327b9dc3e118133cefb1f8d328c22f27786/vip_hci/preproc/cosmetics.py#L156

### problem

The documentation does not match the behaviour.

``` python
n = 3
m = 5
arr = np.array([ 1,  2,  3,  4,  5,  6,  7,  8,  9, 10])

arr[n+1:m+1]
# → array([5, 6])
```

This means the `n+1`th element (0-indexed, or `n+2`nd element in 1-indexed) is the first to be kept.

With `n` and `m` from above, I would expect the output to be

``` python
array([4, 5, 6])
```

(or even `array([3, 4, 5])` if we would like 1-indexed behaviour, which I wouldn't prefer)

### solution

Line 156 should be changed to

``` python
array_view = array[n:m+1, :, :].copy()
```

r4lv:
closes #330

### DIFF
--- a/vip_hci/preproc/cosmetics.py
+++ b/vip_hci/preproc/cosmetics.py
@@ -1,8 +1,8 @@
 #! /usr/bin/env python
 
 """
-Module with cosmetics procedures. Contains the function for bad pixel fixing. 
-Also functions for cropping cubes. 
+Module with cosmetics procedures. Contains the function for bad pixel fixing.
+Also functions for cropping cubes.
 """
 
 from __future__ import division, print_function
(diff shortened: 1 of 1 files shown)


r4lv:
The behaviour now matches your comment from https://github.com/vortex-exoplanet/VIP/issues/330#issuecomment-431867597.

PR #332, opened 2018-10-15
FIX: cube_drop_frames returns tuple

r4lv:
`vip.preproc.cosmetics.cube_drop_frames` returns a tuple `cube, angles`.


carlgogo:
Good but there is still a problem. The HCIDataset can be instantiated w/o the angles, in this case ``self.angles = None`` and ``cube_drop_frames`` has a single output (not a tuple).

r4lv:
BTW, would it make sense to `.drop_frames` in a cube without angles? In what cases would that be used?

PR #335, opened 2018-10-15
ENH: don’t force .npz when loading HCIDataset

r4lv:
We faced something similar with `open_fits` already, now this PR extends the support for files *without* a file extension to HCIDataset's `.load`.

This is especially useful when using `astropy.utils.data.download_file` (which creates files without file extension), e.g. in tests.

### DIFF
--- a/vip_hci/conf/utils_conf.py
+++ b/vip_hci/conf/utils_conf.py
@@ -70,10 +70,10 @@ class Saveable(object):
 
     @classmethod
     def load(cls, filename):
-        if not filename.endswith(".npz"):
-            filename += ".npz"
-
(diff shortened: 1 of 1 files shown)


vip_hci/conf/utils_conf.py
```diff
@@ -72,7 +72,7 @@ def save(self, filename):
     def load(cls, filename):
         try:
             data = np.load(filename)
-        except FileNotFoundError:
+        except:
```
r4lv: `np.load` also raises `EOFError: Ran out of input` and `OSError: Failed to interpret file '...' as a pickle`.

PR #336, opened 2018-10-16
FIX: full_output caused ADI+mSDI PCA to break 

r4lv:
HCIPca calls the pca submodule with `full_output=True`, which caused several errors. I walked through the code and removed the `full_output` where it was actually never needed.


vip_hci/pca/pca_fullfr.py
```diff
@@ -583,6 +641,8 @@ def _adi_rdi_pca(cube, cube_ref, angle_list, ncomp, scaling, mask_center_px,
         pcs = reshape_matrix(V, y, x)
         recon = reshape_matrix(reconstructed, y, x)
     else:
+        # TODO: pcs and recon are not defined when full_output=False, but still
+        #       returned.
```
r4lv: I will also open an Issue for this 👍

PR #411, opened 2019-10-16
Debug of mcmc routine and cube_planet_free, and modified full output for 2d fits

VChristiaens:
Major changes:
- debug of negfc/mcmc_negfc_sampling(): description says that the annulus width and aperture radius should be provided in pixels. In practice the high level code of mcmc_negfc_sampling() assumes annulus_width is indeed in pixels, but that both annulus_width and aperture radius are in FWHM when called in the mcmc sampler, which makes use of PCA-annulus.
- debug of utils_negfc/cube_planet_free(): works fine for a 3d cube, but not for a 4d (ifs+adi) cube. In the latter case, one needs to use a different flux at each wavelength, hence r, theta and flux are all 1d array. Since input cube is 4d, it triggers the ifs mode of cube_inject_companions(), but given the loop on both r and flux, which are both n_wavelength dimension, the negative fake companion is injected n_wavelength x n_wavelength times (i.e. n_wavelength times too many).
- in the case of fits to 2d gaussian, moffat or airy, it is often desirable to have an idea of the uncertainty on the best-fit parameters (in particular on the centroid). The full_output keyword now adds additional entries in the output dictionary, including errors on the best-fit parameters.
- several other minor bugs, including making contrast_curve() compatible with an input fwhm that is 1d array or list. Note that a proper debugging of the contrast_curve() function in the case of a 4d input cube is still required.

### DIFF
--- a/vip_hci/conf/utils_conf.py
+++ b/vip_hci/conf/utils_conf.py
@@ -1,4 +1,5 @@
 #! /usr/bin/env python
+# -*- coding: utf-8 -*-
 
 """
 Module with utilities.
--- a/vip_hci/metrics/contrcurve.py
+++ b/vip_hci/metrics/contrcurve.py
@@ -496,7 +496,10 @@ def throughput(cube, angle_list, psf_template, fwhm, pxscale, algo, nbranch=1,
             else:
                 if algo_dict['scale_list'].shape[0] != array.shape[0]:
                     raise TypeError('Input wavelength vector has wrong length')
-                maxfcsep = int((array.shape[2] / 2.) / fwhm) - 1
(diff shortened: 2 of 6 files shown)


carlgogo:
Hi @VaChristiaens, this is very helpful. The next step is debugging and making sure things won't break in the future, that is test automation. Therefore it would be great to create test functions (at the root level there is a ``tests`` folder) for both the negfc and the contrast curve procedures. These are critical components of ``VIP`` that are used by many of you, but are still not covered by pytest. See here: https://codecov.io/gh/vortex-exoplanet/VIP/tree/master/vip_hci.

I'll merge this PR now.

PR #421, opened 2020-03-09
PCA-ASDI contrast curve bug fixed + other minor bugs corrected

VChristiaens:
- PCA-ASDI contrast curves were not reliable due to a combination of (1) enforcing the scaling in the PCA functions and (2) the flux normalization before the contrast curve calculation not being properly taken into account due to the scaling.  This is now solved by not enforcing the scaling to be used in the PCA function and by computing noise levels in the final images obtained with and without scaling in the contrast curve function.

- Bugs related to python 2 syntaxes still present in the MCMC_smapling and nested_sampling modules are now corrected. 

- Bugs related to the way the aperture and annulus sizes are defined for NEGFC are also corrected. Now the definition is uniform throughout all NEGFC functions (simplex and MCMC), match the description, and propagates well down to the PCA function used in the the NEGFC core function. The input radius is now to be expressed in terms of FWHM, while the annulus width should be in pixels (this is for consistence with the input format of the PCA_annular function). A new option was added to be able to use PCA_annular (i.e. with a PA threshold) instead of PCA_annulus with NEGFC.

- Bad pixel correction algorithm cube_fix_badpix_isolated() is now adapted to work better, in particular to correct for cosmic rays throughout the cube. The bad pixel identification was previously based on the mean frame of the cube, hence cosmic rays were not picked up. I have now added an option to do the bpix correction frame by frame.

- Finally the get_square() function is now compatible with a rectangular input array.

### DIFF
--- a/vip_hci/metrics/contrcurve.py
+++ b/vip_hci/metrics/contrcurve.py
@@ -151,14 +151,14 @@ def contrast_curve(cube, angle_list, psf_template, fwhm, pxscale, starphot,
     if transmission is not None:
         if not isinstance(transmission, tuple) or not len(transmission) == 2:
             raise TypeError('transmission must be a tuple with 2 1d vectors')
-    if isinstance(starphot, float) or isinstance(starphot, int):
-        pass
-    else:
-        if starphot.shape[0] != cube.shape[0]:
-            raise TypeError('Correction vector has bad size')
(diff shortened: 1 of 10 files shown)


VChristiaens:
- other additional option: name and directory of the output files for the NEGFC/MCMC.

PR #430, opened 2020-04-27
Problem with max distances in cube_inject_companions 

GillesOrban:
See #429 for previous discussion.
It turns out that only modifying `noise_per_annulus` is required to makes sure we are not too close from the edge of a frames, where "too close" is here 2*separation and the given separation is `fwhm_med` in `throughput`.

As commented in the function: The annuli start at init_rad (== fwhm by default) and stop 2 x separation before the edge of the frame.
Factor 2*sep is ensured by the -1 line 805 and by the range() function in the for-loop.

### DIFF
--- a/vip_hci/metrics/contrcurve.py
+++ b/vip_hci/metrics/contrcurve.py
@@ -804,7 +804,7 @@ def noise_per_annulus(array, separation, fwhm, init_rad=None, wedge=(0, 360),
 
     if init_rad is None:
         init_rad = fwhm
-
+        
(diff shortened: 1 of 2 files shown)


ChrisDelaX:
I understand this problem was fixed after v0.9.11 which is the latest version of VIP.
It would be nice to have a new version of VIP with the fix included so that I can update the HEEPS dependencies (currently VIP v0.9.9 only doesn't crash HEEPS for demo notebooks).

PR #433, opened 2020-08-06
Update setup.py

jmilou:
Update to make it compatible with the latest pip release (>=20.1.1) which generates an error while installing vip_hci:

The command python setup.py develop
returned the following error:

Traceback (most recent call last):
  File "setup.py", line 40, in <module>
    reqs = [str(ir.req) for ir in reqs]
  File "setup.py", line 40, in <listcomp>
    reqs = [str(ir.req) for ir in reqs]
AttributeError: 'ParsedRequirement' object has no attribute 'req'

The 'ParsedRequirement' object has now an attribute called requirement. The proposed change is retro-compatible.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,10 +37,15 @@ def resource(*args):
 
 # parse_requirements() returns generator of pip.req.InstallRequirement objects
 reqs = parse_requirements(resource('requirements.txt'), session=False)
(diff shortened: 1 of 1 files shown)


VChristiaens:
Thanks @jmilou!

PR #452, opened 2021-05-19
Bug fix for skimage version => 0.18 and improvement in fake disk models 

jmilou:
This pull request suggest 2 changes:
- a change to fix a problem with version of skimage >= 0.18 where skimage.measure.compare_ssim has disappeared and was replaced by skimage.metrics.structural_similarity
- a minor improvement for dust modelling

### DIFF
--- a/vip_hci/pca/pca_fullfr.py
+++ b/vip_hci/pca/pca_fullfr.py
@@ -744,19 +744,19 @@ def _adi_rdi_pca(cube, cube_ref, angle_list, ncomp, scaling, mask_center_px,
     """ Handles the ADI+RDI post-processing.
     """
     n, y, x = cube.shape
-
+    n_ref, y_ref, x_ref = cube_ref.shape
     angle_list = check_pa_vector(angle_list)
     if not isinstance(ncomp, int):
         raise TypeError("`ncomp` must be an int in the ADI+RDI case")
-    if ncomp > n:
-        ncomp = min(ncomp, n)
+    if ncomp > n_ref:
+        ncomp = min(ncomp, n_ref)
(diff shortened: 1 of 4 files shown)


vip_hci/pca/pca_fullfr.py
```diff
@@ -774,19 +774,19 @@ def _adi_rdi_pca(cube, cube_ref, angle_list, ncomp, scaling, mask_center_px,
...
+    n_ref, y_ref, x_ref = cube_ref.shape
     angle_list = check_pa_vector(angle_list)
     if not isinstance(ncomp, int):
         raise TypeError("`ncomp` must be an int in the ADI+RDI case")
-    if ncomp > n:
-        ncomp = min(ncomp, n)
+    if ncomp > n_ref:
+        ncomp = min(ncomp, n_ref)
         msg = 'Number of PCs too high (max PCs={}), using {} PCs instead.'
         print(msg.format(n, ncomp))
```
VChristiaens, line 784: Thanks for spotting this @jmilou. Could you also replace `n` by `n_ref` at line 784? Thanks!

VChristiaens (changes requested):
Thanks @jmilou! The changes look good. I had also noticed the issue regarding skimage version. For the next major VIP update, I was thinking to update the corresponding requirement for VIP to be 'skimage >= 0.18', such that the try/except statement would not be necessary anymore. In the meantime, what you propose should be fine.

vip_hci/pca/pca_fullfr.py
VChristiaens: resolved in following PR (#453)

PR #453, opened 2021-05-19
Improvement in fake disk functions 

jmilou:
This change improves the fake disk models in VIP, by adding information on the location of the max dust density location.


jmilou:
Update after @VChristiaens comment regarding the RDI correction

PR #454, opened 2021-05-19
Bug fix PCA RDI 

Na1an:
The maximum number of PCA components in PCA RDI is currently set to the number of frames of the science cube. This should be set to the number of frames of the reference cube.

### DIFF
--- a/vip_hci/pca/pca_fullfr.py
+++ b/vip_hci/pca/pca_fullfr.py
@@ -780,9 +780,9 @@ def _adi_rdi_pca(cube, cube_ref, angle_list, ncomp, scaling, mask_center_px,
     if not isinstance(ncomp, int):
         raise TypeError("`ncomp` must be an int in the ADI+RDI case")
     if ncomp > n_ref:
-        msg = 'Number of PCs too high (max PCs={}), using {} PCs, the number frames of cube reference, instead.'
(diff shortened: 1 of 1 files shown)


vip_hci/pca/pca_fullfr.py
```diff
@@ -774,14 +774,15 @@ def _adi_rdi_pca(cube, cube_ref, angle_list, ncomp, scaling, mask_center_px,
...
+
     if not isinstance(ncomp, int):
         raise TypeError("`ncomp` must be an int in the ADI+RDI case")
-    if ncomp > n:
-        ncomp = min(ncomp, n)
-        msg = 'Number of PCs too high (max PCs={}), using {} PCs instead.'
-        print(msg.format(n, ncomp))
+    if ncomp > n_ref:
+        msg = 'Number of PCs too high (max PCs={}), using {} PCs, the number frames of the reference cube, instead.'
+        print(msg.format(ncomp,n_ref))
```
VChristiaens, line 784: Thanks for noticing this. I saw a similar fix included in @jmilou's pull request that came just before yours.
I think that line 783 should be rephrased to make more sense, since max PCs is what should be used when the requested ncomp is larger than n_ref (i.e. currently the 2 fields should have the same value in this sentence).  Probably better to have:
`msg = 'Requested number of pcs ({}) higher than the number of frames in the reference cube ({}); using the latter instead.' `

VChristiaens:
Fixed in PR #453

PR #534, opened 2022-05-02
Adding modification of the test function 

kanti92:
Hi Valentin. I have modified the test function, so now it works with both even and odds sizes.


kanti92:
Solving error on test function

kanti92:
I saw the error. I modified the get_circle function. Now it also has the option to return indices

kanti92:
Now, I have modified the truth_even matrix. It was wrong.

VChristiaens:
Thanks @kanti92 !

PR #536, opened 2022-05-03
PACO Algorithm into VIP

nenasedk:
Adding in PACO algorithms

Includes:

- new paco.py file in invprob
- Updated psf subtraction notebook to include basic PACO tutorial
- test coverage of FastPACO
- Updated JOSS paper to include reference to PACO
- Improvements to PACO: use of VIP functions, correcting outputs of FluxPACO, improving documentation

PACO is based on Flasseur+ 2018. This implementation includes alg. 1 (FullPACO), alg2 (FastPACO) and alg 3 (Flux estimation) from the paper.


nenasedk:
Looks like there was a problem detecting the signal - the tests worked ok on my end, but I had to change the orientation of the outputs to be consistent with the tests, so I wonder if that change is what broke things. I'll take a look and update the pull request.

VChristiaens:
Ok thanks for having another look. Strangely enough it looks like `fast_paco_parallel` worked - it is just `fast_paco` that did not recover the fake companion.

nenasedk:
Done. The single core had different index ordering than the parallel outputs, so the results were transposed. Fixed.

PR #538, opened 2022-05-09
Optimized NEGFC and FMMF tests + compacted code for clip_array routine

VChristiaens:


IainHammond:
changes to clip_sigma code seem to have broken it. will try to spot a fix 
image

PR #556, opened 2022-10-21
subtract radial profile and recentering multiprocessing

IainHammond:
hello here are some small new features 
- (stats.imstats) frame_average_radprofile can now output a radial average subtracted frame. False by default to avoid bugs with exisiting codes and pipelines. Masks the inner radius of subtracted frame if init_rad is provided. Added missing info to docstring
- (preproc.recentering) changed default nproc from 1 to None. In the case of nproc=None, it is set to half the number of available CPUs. This is to ensure consistent behaviour for each function and take advantage of some multiprocessing if a user does not provide an explicit input
- (preproc.recentering) nproc has been provided to each call to cube_shift, enabling a small reduction in processing time when using multiprocessing 
- (misc) some general typo fixes and clean up


VChristiaens:
Thanks!

PR #559, opened 2022-11-26
corrected several docstrings

IainHammond:
updated docstrings in fm and bad pixel correction modules to reflect new, changed or removed variables


VChristiaens:
Thanks @IainHammond !

PR #565, opened 2022-12-08
bug fix in cube_correct_nan

IainHammond:
fix for array being saved as object rather than float, only occurs in multiprocessing. fix tested on naco data.


VChristiaens:
Thanks!

PR #567, opened 2022-12-14
misc. fixes

IainHammond:
- fix for array slicing with the good pixel mask with numba in ` _clip_array`
- fixed error when mode='circle' in `cube_detect_badfr_pxstats`
- `detection` returns snr of sources above snr threshold when full_output=True as in docstring


VChristiaens:
Thanks!

PR #579, opened 2023-03-08
Tests updated after a correction in the class ScatteredLight

jmilou:
I reviewed the last test that failed in test_fm_scatteredlight and updated it


VChristiaens:
Thanks @jmilou! All tests have now passed.

PR #581, opened 2023-03-15
Depreciation issues

arionaris:
One issue on setup.py similar to : 
https://stackoverflow.com/questions/62114945/attributeerror-parsedrequirement-object-has-no-attribute-req

And a depreciation issue on recentering.py due to inhomogeneous array unsupported by np.array without specifications.
This was causing a bug on  notebook https://github.com/vortex-exoplanet/VIP_extras/blob/master/tutorials/02_preproc.ipynb
On the 2.1.3 and 2.1.4 sections.


VChristiaens:
Thanks a lot @arionaris!

PR #585, opened 2023-04-01
Refactoring oop upgrades

Eixalde:
- Added lots of support centered around objects (parameters management, object reusability, saving results, etc)
- Regrouped all objects in a sub-folder 'objects'
- Added tests for all of that new content
- Included in-code typing documentation

PCA is still unstable and will be fixed in the very next branch, until then caution is advised when using its object


Eixalde:
I don't really understand why the previous tests were cancelled, although it seems they work with 3.10 without any 3.10 support yet. I think it is safe to assume the tests would have been passed if not cancelled

VChristiaens:
Thanks a lot @Eixalde!

PR #587, opened 2023-05-04
Fix various issues

Eixalde:


Eixalde:
Problems fixed so far :
- codecov dependancy update
- URL requests for test resources now have cooldown if necessary

Problems to be considered (after merge) :
- docs not generating properly

PR #604, opened 2023-06-15
Remove strenum optional dependancy

Eixalde:
Replaced strenum by generic, Python-included Enum support. Works just like before


VChristiaens:
Thanks @Eixalde !

PR #622, opened 2023-10-25
fixed VIP installation problems related to scikit-image, general improvements for recenter functions and docstrings , PEP8 imports

IainHammond:
- VIP was facing installation difficulties with Python 3.9, 3.10 and 3.11 due to requiring a very old version of scikit-image (0.18.3). The limitation was due to a new parameter `normalization` being added to `phase_cross_correlation`, being set to true by default, which changed the resulting shifts. We now provide `normalization=None` to retain the original behaviour. _shift_dft also catches the case where the latest version of `phase_cross_correlation` (0.22) from scikit-image returns all three variables always rather than just the shifts (which the VIP code was originally designed around), and keeps the original behaviour of the _shift_dft function. The code is now compatible with all versions of scikit-image , however I recommend 0.22 (latest release) due to the performance improvements
- in cube_recenter_speckles, the cube is only high pass filtered if the user requests it (by default it is), and I've moved each call of .copy() to numpy.copy() as it's safer and faster
- avoided a photutils import deprecation
- PEP8 a bunch of imports and removed some unused ones
- docstring fix in cube_recenter_dft_upsampling related to the center_fr1 parameter
- changed print statements to be consistent with "FITS"
- general typo fixes


VChristiaens:
Thanks @IainHammond !

PR #625, opened 2023-11-21
Correction to how the reference dataset is built in do_pca_patch

Julat300:
In the function "do_pca_patch" used in the PCA_annular algorithm, in some cases, the reference images (matrix_ref) were stacked twice in data_ref in the RDI+ADI case


VChristiaens:
Thanks @Julat300 !

PR #638, opened 2024-04-11
DFT shift fix for new scikit, speed improvement for find_nearest

IainHammond:
This request is mainly to address the upgrade from scikit-image 0.18.3 to 0.23.1, which released today. 
Using `phase_cross_correlation` in the `_shift_dft` function would not give a backwards compatible answer for versions>0.18.3 without providing `normalization=None`. There is now a check on the version of scikit to handle the case where a user has scikit 0.18.3 and this parameter doesn't exist, and two outputs are returned from the function (compared to three in 0.22 ->). The version requirement has been removed from requirements.txt and a standard installation of VIP will grab scikit-image 0.23.1. 

I've also sped up find_nearest with some boolean operations and numpy functions (not that it needed it).

fakecomp.py and fit_2d.py have had a small tidy up relating to the import for centroid_com from photutils.centroids as the old method for importing this function is 8 years old. Avoiding this check makes the initial start up faster by avoiding the entire package being imported, which can be about 160MB


VChristiaens:
Thanks @IainHammond !

PR #644, opened 2024-08-09
more robust mu sigma calculation, new seeing function, memory improvements, typo fixes

IainHammond:
bit of a mixed bag for this one. the main changes are:
- `get_mu_and_sigma` now ensures the guess of a planet separation is larger than 1xFWHM
- `get_mu_and_sigma`: `radius_int` can't be negative anymore
- `get_mu_and_sigma`: calculation for mu and sigma uses an annulus that is better within the vicinity of the planet separation
- new function alert: `seeing_from_header` very simple and extremely fast function to measure the seeing over a list of FITS files (e.g. a night of observations). Could be expanded to include wind speed.
- `derotation.py`: np.nan_to_num is performed on the output array rather than making a copy of it 
- `get_snr`: the snr vs. npc plot now only plots integer values for npc on the x axis (no more 1.5 npc etc.)
- `dist`: function now just uses np.hypot
- `bad_frames.py`: moved all imports to the top of file
- `bad_frames.py`: DAOStarFinder is now a lazy import. This was by far the slowest and generally added 2seconds to an initial import of any VIP function (benchmarked with `tuna`)


VChristiaens:
Thanks @IainHammond !

PR #651, opened 2024-11-06
Torch FFT Rotation & Juillard23 IPCA

Sand-jrd:
**1-Adding Torch for Faster FFT Rotation Mode**
Fft rotation using `vip-fft` mode use `Numpy` library. The use of `Torch` instead of `Numpy` consistently improves speed. New `imlib='torch-fft'`.Having Torch remains an optional dependency, and a check to determine whether or not Torch is installed has been implemented.

Here are the differences between the Torch FFT and VIP FFT methods for 50 rotations of a cube. The time is measured in seconds. The third image shows the difference between both rotation methods. The use of `Torch` instead of `Numpy` consistently improves speed.
torch fft 2 780

**FYI (NOT IN THIS COMMIT)- Torch with Built-in Rotation Method (Bilinear/Nearest) not faster than Open CV**
Here are the differences between the Torch and OpenCV methods for 50 rotations of a cube, with time measured in seconds. The third image shows the difference between the two rotation methods. In this case, the use of OpenCV is better, hence, I did not included  the option of using Torch Built-in Rotation Method in the commit.
Pytorch 0 269

**2-Adding “Juillard23” Mode to IPCA**
The `“Juillard23”` mode has been added to the IPCA mode, enabling users to utilize the implementation from Juillard et al. (2023/2024). This method has no additional options for significant signal extraction and works exclusively with Torch, making it faster but also more prone to propagating noise and disk flux. The mode can return a stim map and residual map if full-output is set to true. This option is only compatible with `'ADI' `and `'ARDI'`; an error is raised if the options do not match. Other parameters, such as imbil and interpolation, are ignored. This option only uses `nit`, `ncomp`, `pup_mask_center_px`, `cube_ref`, `cube`, and `angle_list`. In particular, the threshold is always set to 0, and STIM is only computed at the end and is not part of the iterative process. Installation of the GreeDS package (via `pip install GreeDS`) is required for this option. GreeDS is now a new optional dependency of VIP.
vip_fft 15 219753 Greeds Sand 0 496539

### DIFF
--- a/vip_hci/greedy/ipca_fullfr.py
+++ b/vip_hci/greedy/ipca_fullfr.py
@@ -34,6 +34,13 @@ from ..preproc import cube_derotate, cube_collapse
 from ..metrics import stim_map, inverse_stim_map
 from ..var import prepare_matrix, mask_circle, frame_filter_lowpass
 
+try:
+    from GreeDS import GreeDS
+    no_greeds = False
+except ImportError:
+    msg = "GreeDS python bindings are missing."
+    warnings.warn(msg, ImportWarning)
+    no_greeds = True
 
 @dataclass
 class IPCA_Params(PCA_Params):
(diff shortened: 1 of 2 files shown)


Sand-jrd:
I need to remove the type definitions to fix the check failures.

VChristiaens:
Great! Thanks a lot @Sand-jrd !

PR #658, opened 2025-01-24
Modern multiprocessing for MCMC NEGFC, small fix to tutorial 05A

IainHammond:
So emcee has been locked into version 2.2.1 for some time, which was release on 15 July 2016. The reason was likely to be due to the change in multiprocessing behaviour in version 3.0 -> with the removal of the `threads` argument in favour of `pool` in `multiprocessing`.

From https://emcee.readthedocs.io/en/stable/user/upgrade/
"_The `threads` keyword argument has been removed in favor of the `pool` interface (see the Parallelization tutorial for more information). The old interface had issues with memory consumption and hanging processes when the sampler object wasn’t explicitly deleted. The `pool` interface has been supported since the first release of emcee and existing code should be easy to update following the Parallelization tutorial._"

This older version was indeed causing some occasional problems (at least for me) such as memory accumulating or the code seemingly stopping. To overcome this I've jumped us up to emcee 3.1.6 and made all of the recommended changes to the call for `emcee.EnsembleSampler` provided in the docs. It works, and it produces the same results. Is it faster? A bit. The user doesn't need to manually assign the number of processors anymore either as python automatically determines what is available. The `multiprocessing` start method is forkserver by default as it was running about 10% faster than fork. If it's unavailable, spawn is used instead so that Windows is supported (albeit spawn is about 20% slower than forkserver). Finally, a new `moves` keyword replaces the `a` keyword, and `sampler.lnprobability` is now `sampler.get_log_prob()`.

The implication of this is that VIP technically needs emcee 3.0 -> from now on and requirements.txt has been updated to remove the version lock. In the long term this is much better, but will cause an error if a user doesnt update emcee following this pull request.

Random things:
- fixed a wrong keyword argument in tutorial 05A
- fixed a deprecated ipython import

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ astropy
 photutils
 scikit-learn
 scikit-image
-emcee==2.2.1
+emcee
 nestle
 corner
 pandas
--- a/tests/pre_3_10/test_preproc_recentering.py
+++ b/tests/pre_3_10/test_preproc_recentering.py
@@ -32,7 +32,7 @@ from vip_hci.var import frame_center, fit_2dgaussian
 mpl.use("Agg")
 
 try:
-    from IPython.core.display import display, HTML
+    from IPython.display import display, HTML
 
     def html(s):
         display(HTML(s))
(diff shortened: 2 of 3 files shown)


VChristiaens:
Thanks for the changes @IainHammond. I think it'd still better for the user to be able to provide the number of CPUs to use for the MCMC. In many scenarios, either the user has to share the machine with others, or may want to run other jobs on their machine simultaneously, so would not want all available CPUs to be used. Can you adapt this?

IainHammond:
providing nproc to Pool again

IainHammond:
ugh one of the tests didnt download the data it seems. It worked on my fork, which is identical

VChristiaens:
Thanks @IainHammond !

PR #659, opened 2025-01-25
New functionalities to vip_hci/fm/scattered_light_disk.py

jmilou:
The pull request adds a new function to scattered_light_disk.py: the possibility to generate a map with the scattering angles in the disk mid plane (useful to see which scattering angles correspond to which part of the disk). 
Another change allows the star to be at a position different from the center of the image .


VChristiaens:
Thanks a lot @jmilou !

PR #662, opened 2025-03-18
Add greedy in setup.py

Sand-jrd:
Greedy was missing in the setup.py package, which causes problems when installing the package via pip.


Sand-jrd:
Note that this is urgent for all releases that introduced IPCA because, without this line, the installation via pip install does not include IPCA.
**However, it is not necessary if the pull request #653 replacing setup.py with pyproject.toml is validated beforehand !**

VChristiaens:
Thanks @Sand-jrd !

PR #670, opened 2025-07-30
BLD: move static metadata to \`pyproject.toml\`, explicitly specify build backend and requirements

neutrinoceros:
This allows building the package from source with arbitrary buid frontends, not just pip.
For instance, this allows:
```
uv build
```
or
```
python -m build
```

having `pip` as a build-time dependency is unsual and could easily be avoided by moving the information currently contained in `requirements.txt` and `requirements-dev.txt` to `pyproject.toml`, but that would also break any existing development workflows explicitly relying on their presence, so I consider it out of scope for now, but please let me know if you'd like me to do it too, either here or as a follow up PR.
I could also add explicit support for Python 3.12 and 3.13, but again, I'll treat it as out of scope unless explicitly agreed on first.


pyproject.toml
```diff
@@ -0,0 +1,33 @@
+[build-system]
+build-backend = "setuptools.build_meta"
+requires = [ "pip", "setuptools>=77" ]
```
neutrinoceros, line 3: note: I've used `setuptools>=77` here to get support for `PEP 639` license specification (`project.license-files`), but unfortunately this version of setuptools isn't available for Python 3.8. Would it be okay to also drop support for this end-of-life version of Python too ?

IainHammond: I'm in support of dropping it but Valentin should make the final call

neutrinoceros:
yeah, I only noticed it wasn't explicitly tested, but I didn't hit any issue with it

neutrinoceros:
Just noticing now that #653 already existed but seems to have stalled.
Happy to add @Sand-jrd as a co-author on my commits if it helps moving this forward

neutrinoceros:
Fun fact: Python 3.7 was dropped exactly 2 years ago today in https://github.com/vortex-exoplanet/VIP/pull/610
Given the precedent, I'm hoping that dropping 3.8 now isn't controversial

neutrinoceros:
Oh, interesting, apparently CI doesn't catch the *known* issue with building on Python 3.8 because it never actually builds the package and instead runs against the source tree. This is also easy to fix, but requires a possibly disrupting change of the repository layout from flat to src. See this doc if this terminology is unfamiliar: https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout

neutrinoceros:
I see a test already failed
```
tests/pre_3_10/test_dataset.py F
```
Unfortunately the same test seems to pass locally (tried on 2 different machines) so I'll just wait for CI to complete to see what's wrong. I also note that I couldn't run the *entire* test suite on either of my machines as it seems prohibitively costly (and I think I'm seeing some issues with parallelism)

neutrinoceros:
nevermind I got one to finish and I see now that the failure is also a symptom of a problem I already hinted at: CI is testing against the source tree, not a built package. Anyway I'll just work around it for this fundamental PR and will fix the workaround later on !

neutrinoceros:
CI finally looks good, thank you for your patience !
note: I might be able to tackle a couple of issues I identified this very afternoon if this one is merged soon enough (but of course, no big deal if it isn't).

PR #671, opened 2025-07-30
DEP: move dependency specs to static metadata

neutrinoceros:
based off #670
I originally said I would only do this if requested but I changed my mind and thought it might be more compelling to show a concrete implementation of my proposal.


pyproject.toml
```diff
@@ -0,0 +1,72 @@
...
+    "pandas",
+    "matplotlib>=3.7.0",
+    "psutil",
+    "pyprind",
+    "munch",
+    "hciplot>=0.2.4",
+    "dataclass_builder",
+]
+
+[project.optional-dependencies]
```
neutrinoceros, line 49: note: this is strictly equivalent to what used to be defined dynamically in `setup.py`, however I believe there's now a better approach with recent standards, using a [PEP 735 dependency group](https://peps.python.org/pep-0735/), but that would require dropping support for Python 3.8 (end of life), because no version of `pip` supports both.

pyproject.toml
```diff
@@ -0,0 +1,72 @@
...
+dev = [
+    "opencv-python",
+    "sphinx",
+    "myst-nb",
+    "myst-parser",
+    "pandoc",
+    "pytest",
+    "pytest-cov ~=2.6.0",
```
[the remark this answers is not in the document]
neutrinoceros: Honestly this plugin should be unnecessary altogether, using coverage by itself is not that much more complicated and reduces the attack surface for dependency-hell to break loose. Happy to implement this migration too as a follow up !

vip_hci/preproc/recentering.py
```diff
@@ -1382,15 +1382,10 @@ def _shift_dft(array_rec, array, frnum, upsample_factor, mask, interpolation,
     scikit-image 0.19. This should be set to None to maintain the original
     behaviour of _shift_dft.
     """
-    if version("scikit-image") > "0.18.3":
```
neutrinoceros, line 1385: I found this branching and wanted to highlight it as buggy because string comparison is not appropriate for version checking. For instance
```py
>>> "10.0.0" > "2.0.0"
False
```
The *correct* way to compare versions would be
```
from packaging.version import Version
from importlib.metadata import version

SCIKIT_IMAGE_VERSION = Version(version("scikit-image")

if SCIKIT_IMAGE_VERSION > Version("0.18.3"):
    ...
else:
	...
```
where `packaging` is a third party package that needs to be declared as a dependency...
but in this instance I find that `scikit` `0.18.3` is old enough that it's actually much simpler to drop support for any version older than `0.19.0`, in particular because *that* version is the first one that supports Python 3.10, so, if we end up dropping 3.9, no older version *could* be installed anyway.

IainHammond: This is fine, I thought this code had already been removed actually.

VChristiaens: Fine by me too to just remove the check for the scikit-image version.

PR #673, opened 2025-08-01
MNT: migrate package layout from flat to src

neutrinoceros:
This was discussed in #670, where we saw multiple issues with testing surfacing that were symptoms of CI running tests again the source tree instead of a built package. Using a "src-layout" makes this situation much less likely to happen. I'm going to do this change in two commits, where the first one will be moving all source files, which is expected to fail CI, and in the second one I'll adapt CI, resolving the core issue.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -7,4 +7,4 @@ coverage:
 
 ignore:
   - tests/*
-  - vip_hci/vip_ds9.py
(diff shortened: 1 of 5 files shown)


neutrinoceros:
Hum, actually, it might also require #671 goes in first. Will keep this one as a draft for now

VChristiaens:
Thanks a lot @neutrinoceros !

PR #711, opened 2026-04-14
critical indexing fix for annular PCA-ASDI

IainHammond:
fixed an indexing mistake for 4D cubes (the wavelength axis was indexed instead of the temporal axis)


VChristiaens:
Thanks for catching this @IainHammond !

PR #712, opened 2026-04-15
Andromeda: fix numpy error due to dividing a boolean array

IainHammond:
I was getting a critical error when running Andromeda with `homogeneous_variance=False`, looks like it was just a problem with modern numpy being strict


VChristiaens:
Thanks for the PR @IainHammond !

PR #720, opened 2026-06-13
Few major bug fixes + improvements to the test suite

IainHammond:
This one is not as big and scary as it looks! 

ANDROMEDA 
- Fixed a bug where the `opt_method` parameter in the function and in paramenum had an incorrectly labelled technique called `robust`, instead of `l1`. Trying to use `robust` would throw an error, so this is now corrected to `l1` everywhere
 - Using faster numpy and scipy functions to replace some of the slower code that gets called many thousands of times

Bad pixel removal
- Dropped an unnecessary double array operation
- Replaced an unnecessary loop with a direct array operation

Crop frames/crop cubes
- Added a check to see if the frame/cube size is already the crop size. If so, the array is simply returned rather than throwing a critical error (this was kind of annoying in scripts and pipelines, or re-running cells in notebooks)

`cube_inject_fakedisk`
- angle_list can now accept a list, as the name implies, to be consistent with the rest of VIP

`pyproject.toml`: general quality of life updates based on PEP standards
- Bumped the version of setuptools to something less ancient lol
- Added URLs
- Explicitly tell setuptools where the the package is (`src`) and tell setuptools that the code has some type annotations
- Tell pytest where the tests are 

`Makefile`
- Fixed a coverage typo

`ci.yml`: decent overhaul of tests
- Fixed a major mistake where pre-commit was not seeing any files at all! I've changed it to use all git-tracked python files, which kicked in the white space and line end of file cleanups, so there are a few files that have been cleaned by the now functioning pre-commit
- Changed coverage to only run on python 3.14 - this is because the coverage is identical on all python versions so we may as well just use it on the most recent (we don't have any paths in the code for different python versions). This speeds up the tests that run without coverage by a few minutes, although it gets slightly absorbed into a randomness of github test durations
- Coverage has been fixed! Previously it was only seeing some fragments of the code and I don't quite understand how it was working, but it wasn't right. Now we save a coverage shard for each of the 3 splits, and then combine and upload one file at the end of the CI (ideally it would be nice to get rid of split tests eventually but they still take a little too long)
- Gave the tests a huge speed boost by balancing them better. They take roughly a similar amount of time now (20-40 mins) and complete faster due to better allocation of resources
- Renewed the .test_durations file with the modern run time of each test and removed all of the old tests that aren't used anymore
- Pinned each actions package to main to avoid them falling far behind like last time
- Caching pip for each of the tests

`py.typed`
- New file (PEP 561) to tell IDE's that VIP has type hints for some functions

Wishlist:
- I want to completely re-do nested sampling using `dynesty` 
- Add better support for ARM-based architecture (like on mac and AMD CPUs)
- (far future) Possibly ditch emcee for gradient based sampling
- (for discussion) move a lightweight, modern version of hciplot to VIP and drop the hciplot package?


IainHammond:
3.13 objects is just being silly

VChristiaens:
Thanks a lot @IainHammond ! Excellent work with all these improvements!

PR #724, opened 2026-07-06
fix for frame_deconvolution

IainHammond:
The best time to submit a pull request is before a new release; the second best time is right after.

A colleague at MPIA wanted to use the frame_deconvolution function in VIP but found it broken. I think it's been a while since it got some love. This was a matter of updating the `num_iter` variable which got renamed a long time ago in skimage, and also opening up the `clip` parameter. Hopefully everything is clean and documented nicely now.

Enjoy!


VChristiaens:
Thanks @IainHammond !

